PR #5, opened 2018-12-20
feature/insight-local-draft

DmitriVanGuard:
This PR makes possible to save locally currently editable insight's:
1. Text;
2. Description;
3. Tags.

After publishing an insight local draft will be removed.


Partysun:
@DmitriVanGuard Where did you save it to localstorage?

DmitriVanGuard:
> @DmitriVanGuard Where did you save it to localstorage?

Yes, it is. The draft is stored inside `user` store, which we save locally [here](https://github.com/santiment/app/blob/12bc5efb0a8551a3eae1e5fa611dec830ae9d769/src/index.js#L61)

PR #20, opened 2018-12-30
feature/trends-graph-volume-score

DmitriVanGuard:
#### Summary
Social volume score chart
#### Screenshots
image


src/components/SocialVolumeScore/SocialVolumeScore.js
```diff
@@ -0,0 +1,290 @@
+import React from 'react'
+import { BarChart, Bar, XAxis, YAxis, ResponsiveContainer } from 'recharts'
+import moment from 'moment'
+import { connect } from 'react-redux'
+import styles from './SocialVolumeScore.module.scss'
+
+const data = [
```
Partysun: @DmitriVanGuard Can you make a story for this, you can use this data.
Anyway, we will have API at the start of the next week. 
It will be good finish this pr with the component and the story with the visual test.

PR #37, opened 2019-01-18
Enhance/new navigation

Partysun:
image


src/components/Navbar/Navbar.module.scss
```diff
@@ -65,16 +80,18 @@
...
     }
   }
 }
 
-.leftLink {
-  margin-right: 4px;
-
-  &_droppable:focus {
-    background-color: $btn-background-light-green;
+@media (min-width: $breakpoint-desktop) {
```
DmitriVanGuard: @Partysun We have a [special mixin](https://github.com/santiment/app/blob/0d4ea5982a75f7afb6766a0f5717a8f1d3145e88/src/mixins.scss#L12) for a responsive design. It can be used like this:
```scss
.wrapper {
/* .... */
   @include responsive('desktop'){
       /* .... */
   }
}
```

PR #39, opened 2019-01-23
enhance/navbar-assets-dropdown-watchlist

DmitriVanGuard:
#### Summary
Update of the NavbarAssetsDropdown Watchlist part.
The current behavior is:
1. Click on the `Create new watchlist` shows the input form;
2. If the input form is empty then onBlur event, `Create new watchlist` will be shown again.

#### Screenshots
image


src/components/Navbar/IconPlus.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+const IconLock = props => (
```
Partysun: @DmitriVanGuard Move icons to san-ui

PR #69, opened 2019-02-07
fix/insight-draft-update-tags

DmitriVanGuard:
#### Summary
Update insight will not cause it to lose the tags.

### DIFF
--- a/src/pages/InsightsNew/CreateTitle.js
+++ b/src/pages/InsightsNew/CreateTitle.js
@@ -33,6 +33,21 @@ const LimitSizeOfTitle = ({ length }) => (
   </small>
 )
 
+const normalizeTags = tags => {
+  return tags.map(({ label, value, name }) => {
+    if (label && value) {
+      return { label, value }
+    }
+
+    if (name) {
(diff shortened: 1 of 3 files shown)


src/pages/InsightsNew/CreateTitle.js
```diff
@@ -33,6 +33,21 @@ const LimitSizeOfTitle = ({ length }) => (
   </small>
 )
 
+const normalizeTags = tags => {
+  return tags.map(({ label, value, name }) => {
+    if (label && value) {
```
Partysun: I see one case when this method returns nothing. When label or value will be null.
@DmitriVanGuard 
Also maybe 
`if (name) {
  return {label: name, value: name}
}
return {label, value}
`

DmitriVanGuard: @Partysun if `label` or `value` will be `null`, then there is a guarantee that tag has a `name` field and the check on the `42`th line will be passed.
Yeah, I will remove the second `if`, it's unnecessary here because of the `return`

PR #73, opened 2019-02-08
fix/social-volume-widget-behaviour

DmitriVanGuard:
#### Summary
- After hovering on trend word, fetching `topicSearch` chart data for every possible source and merging it into one array;

Also:
- A lot of junk code was removed;
- Epic logic was simplified;
- Component now works much faster.
#### Screenshots
image
image


src/components/SocialVolumeWidget/socialVolumeEpic.js
```diff
@@ -1,17 +1,10 @@
 import { Observable } from 'rxjs'
 import moment from 'moment'
 import * as actions from './actions'
-import {
-  socialVolumeGQL,
-  totalSocialVolumeGQL,
-  wordTrendScoreGQL,
-  allProjetsGQL
-} from './socialVolumeGQL'
+import { socialVolumeGQL } from './socialVolumeGQL'
```
Partysun: Let's try to use SOCIAL_VOLUME_QUERY name, as Apollo team does. I used it for new code and tried to refactor old. @DmitriVanGuard

PR #106, opened 2019-03-11
feature/insights-sort

DmitriVanGuard:
#### Summary
Sorting of the insights feed. Sorting logic is done via `sortReducer`.
#### Screenshots
image


src/components/Select/Select.js
```diff
@@ -1,9 +1,17 @@
 import React from 'react'
 import LibSelect from 'react-virtualized-select'
-import './Select.module.scss'
+import cx from 'classnames'
+import './Select.scss'
+import styles from './Select.module.scss'
 
-const Select = ({ ...props }) => {
-  return <LibSelect {...props} optionHeight={32} />
+const Select = ({ className, topDropdown, ...props }) => {
```
Partysun: One more select in our source. Can we use wrapper in san-ui? @DmitriVanGuard

DmitriVanGuard: @Partysun Sure! I will move it out today

PR #111, opened 2019-03-14
feature/insights-seo-links

DmitriVanGuard:
#### Summary
Accessing insights is now possible via SEO friendly link: `/read/title-of-the-insight-:id`. 
This feature is backward compatible with the old link scheme: `/read/:id`

### DIFF
--- a/src/pages/Insights/utils.js
+++ b/src/pages/Insights/utils.js
@@ -43,3 +43,12 @@ export const getInsightContent = htmlContent => {
   tempHTMLElement = null
   return content
 }
+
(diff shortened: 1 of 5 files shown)


src/pages/Insights/utils.js
```diff
@@ -43,3 +43,12 @@ export const getInsightContent = htmlContent => {
   tempHTMLElement = null
   return content
 }
+
+export const getInsightIdFromSEOLink = link =>
+  +link.slice(link.lastIndexOf('-') + 1)
+
```
Partysun: Add small specs for this.

PR #118, opened 2019-03-20
feature/chart-builder

DmitriVanGuard:
#### Summary
Chart builder widget with a shareable state via a link. 

#### Screenshots
- While building
image

- Shared link (view only)
image

### DIFF
--- a/src/ducks/SANCharts/ChartPage.js
+++ b/src/ducks/SANCharts/ChartPage.js
@@ -1,5 +1,6 @@
 import React, { Component } from 'react'
 import * as qs from 'query-string'
+import Loadable from 'react-loadable'
 import GetTimeSeries from '../../ducks/GetTimeSeries/GetTimeSeries'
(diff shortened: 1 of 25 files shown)


src/ducks/SANCharts/ChartPage.js
```diff
@@ -23,30 +35,98 @@ class ChartPage extends Component {
...
+
+  onMetricsChange = metrics => {
+    this.setState({ metrics }, this.updateSearchQuery)
+  }
+
+  onNightModeSelect = () => {
+    const { nightMode } = this.state
+
+    this.setState(
+      { nightMode: nightMode ? undefined : true },
```
Partysun: @DmitriVanGuard Maybe smth diff?

DmitriVanGuard: I agree it looks a little bit ugly and straightforward, but it is the easiest way to hide it from the querystring.

PR #122, opened 2019-03-22
feature/notification

devanshu-santiment:
This PR replaces the Notification component in the app.

Changes:
   - Action `showNotification` now accepts an object with title, description and variant (default variant set to 'info')
   - Notification component now contains the logic to manage notification stack

Screenshot contains updated notification UI
image


src/components/Notification/index.js
```diff
@@ -0,0 +1,83 @@
+import React, { Component } from 'react'
+import { connect } from 'react-redux'
+import { Notification as NotificationItem } from '@santiment-network/ui'
+import SlideEntry from './Animated/SlideEntry'
```
DmitriVanGuard: image

src/components/Notification/index.js
```diff
@@ -0,0 +1,83 @@
...
+import styles from './Notification.module.scss'
+
+class Notification extends Component {
```
DmitriVanGuard: The name is semantically incorrect. It implies the existence of only **one** notification. A more appropriate name would be `NotificationStack`

src/components/Notification/index.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  closeNotification = notificationId => {
+    this.setState(
+      {
+        dissmisedNotification: [
+          ...this.state.dissmisedNotification,
+          notificationId
+        ]
```
DmitriVanGuard: After every dismissed notification handling this garbage through the app lifetime

src/components/Notification/index.js
```diff
@@ -0,0 +1,83 @@
...
+    const { notifications, dissmisedNotification } = this.state
+
+    if (!notifications.length) {
+      return null
+    }
+
+    return (
+      <div className={styles.notificationStack}>
+        {notifications.map(notification => (
+          <SlideEntry leaving={dissmisedNotification.includes(notification.id)}>
```
DmitriVanGuard: The absence of the `key` prop in the array element stack

src/components/Animated/SlideEntry.js
```diff
@@ -0,0 +1,39 @@
...
+
+class SlideEntry extends Component {
+  constructor (props) {
+    super(props)
+
+    this.state = {
+      entered: false
+    }
+
+    setTimeout(() => {
```
DmitriVanGuard: 1. An uncleared timer will cause to the dangling pointer and change the state of the unmounted component.
2. No need to create this component. There is a great [library](https://github.com/reactjs/react-transition-group) (reliable and well documented) for custom animations. Their examples contain scenarios similar to ours

Partysun: About 2. @DmitriVanGuard We have animate-component library inside our app. Or better switch to https://popmotion.io/pose/ (Did you see it).
But for small interaction, who knows what is better.

Partysun: About timer yes, it's a warning.

DmitriVanGuard: Speaking in the context of moving to our custom animations:
1. [react-transition-group](https://bundlephobia.com/result?p=react-transition-group@2.7.0)(4.8 kB) is ~5x smaller gzipped than [react-pose](https://bundlephobia.com/result?p=react-pose@4.0.8)(20.8 kB);
2. `react-transition-group` is recommended in the React [documentation](https://reactjs.org/docs/animation.html);
3. `react-pose` really fits for `styled-components` meanwhile we are using css modules;

Partysun: Ok. Let's use react-transition-group

src/components/Notification/index.js
```diff
@@ -0,0 +1,86 @@
...
+        {notifications.map(notification => (
+          <SlideEntry
+            key={notification.id}
+            leaving={dissmisedNotification.includes(notification.id)}
+          >
+            <NotificationItem
+              {...notification}
+              className={styles.notification}
+              onClose={() => this.closeNotification(notification.key)}
+              solidFill
```
DmitriVanGuard: What is the reason to have it as a bool prop if it is always true? `solidFill` should be specified inside action creator

src/components/Notification/index.js
```diff
@@ -0,0 +1,86 @@
...
+            />
+          </SlideEntry>
+        ))}
+      </div>
+    )
+  }
+}
+
+const mapStateToProps = state => {
```
DmitriVanGuard: Object destruction

src/reducers/notification.js
```diff
@@ -6,8 +6,11 @@ export const initialState = {
 
 const defaultNotificationState = {
```
DmitriVanGuard: Absolutely no need for this object. `id`, `title` are specified in the action creator, other fields are optional and could be applied directly without runtime overhead for `...defaultNotificationState`

src/actions/rootActions.js
```diff
@@ -4,15 +4,27 @@ import {
...
+const defaultNotification = {
+  title: 'Empty message'
+}
+
+export const showNotification = (payload = defaultNotification) => {
   const newPayload =
-    typeof payload === 'string' ? { message: payload } : payload
+    typeof payload === 'string'
+      ? {
+        ...defaultNotification,
```
DmitriVanGuard: Just why?

src/actions/rootActions.js
```diff
@@ -4,15 +4,21 @@ import {
...
+  newPayload.dismissAfter = newPayload.dismissAfter || 4000
 
   return {
     type: SHOW_NOTIFICATION,
     payload: {
       ...newPayload,
-      key: new Date().getTime()
+      solidFill: true,
+      key: notificationId,
+      id: notificationId
```
DmitriVanGuard: 1. Why have both `key` and `id`? (Data duplication)
2. `solidFill`, `dismissAfter` and other defaults should be applied in the reducer.
```js
case SHOW_NOTIFICATION:
return {
   dismissAfter: 4000,
   solidFill: true,
   variant: 'info',
   ...action.payload
}
```

src/components/NotificationStack/index.js
```diff
@@ -0,0 +1,75 @@
...
+
+  state = {
+    notifications: []
+  }
+
+  static getDerivedStateFromProps ({ notification }, prevState) {
+    return {
+      notifications: notification
+        ? [...prevState.notifications, notification]
+        : []
```
DmitriVanGuard: Should return `null` if the state should not be changed.

src/components/NotificationStack/index.js
```diff
@@ -0,0 +1,75 @@
...
+    })
+  }
+
+  componentDidUpdate (prevProps, prevState) {
+    if (this.state.notifications.length > prevState.notifications.length) {
+      const { id, dismissAfter } = this.state.notifications[
+        this.state.notifications.length - 1
+      ]
+
+      this.timerHandles[id] = setTimeout(() => {
```
DmitriVanGuard: Much easier to save last notification into the static variable than doing line `33`-`36`

src/components/NotificationStack/index.js
devanshu-santiment: This won't have any effect as in our case, `getDerivedStateFromProps` only invokes when we have a notification props since we don't have any action like HIDE_NOTIFICATION that can unset the value of notification state

PR #137, opened 2019-03-28
Hotfix/notification animation

devanshu-santiment:
This PR contains fixes for notification component
 key points: 
1. stylefmt won't listen to stylelint's ignore flags https://github.com/morishitter/stylefmt/issues/240
2. Moved notifications to store, makes it easier to manage when we have single source of truth

### DIFF
--- a/src/components/NotificationStack/index.js
+++ b/src/components/NotificationStack/index.js
@@ -3,7 +3,7 @@ import { connect } from 'react-redux'
 import { bindActionCreators } from 'redux'
 import { Notification as NotificationItem } from '@santiment-network/ui'
 import { CSSTransition, TransitionGroup } from 'react-transition-group'
-import * as Actions from '../../actions/rootActions'
+import * as actions from '../../actions/rootActions'
 import styles from './NotificationStack.module.scss'
 
 const notifyDuration = +styles.notifyduration
@@ -32,8 +32,8 @@ class NotificationStack extends Component {
     }
   }
 
-  closeNotification = notificationId => {
-    this.props.hideNotification(notificationId)
+  closeNotification = id => {
+    this.props.hideNotification(id)
   }
 
   render () {
@@ -65,7 +65,7 @@ const mapStateToProps = ({ notification }) => ({
 const mapDispatchToProps = dispatch =>
   bindActionCreators(
     {
-      hideNotification: Actions.hideNotification
+      hideNotification: actions.hideNotification
     },
     dispatch
   )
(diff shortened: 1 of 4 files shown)


src/components/NotificationStack/index.js
```diff
@@ -13,16 +15,6 @@ class NotificationStack extends Component {
     notifications: []
```
Partysun: Why do we need to have this? @devanshu-santiment

src/components/NotificationStack/index.js
```diff
@@ -43,13 +37,11 @@ class NotificationStack extends Component {
...
-    }))
+    this.props.hideNotification(notificationId)
   }
 
   render () {
-    const { notifications } = this.state
+    const { notifications } = this.props
 
     return (
       <TransitionGroup className={styles.notificationStack}>
```
Partysun: Why do we need to have var "i", also we used the full name for this variable ("index") @devanshu-santiment

src/components/NotificationStack/index.js
devanshu-santiment: sorry, pushed a fix

src/components/NotificationStack/index.js
```diff
@@ -43,13 +37,11 @@ class NotificationStack extends Component {
   }
 
   closeNotification = notificationId => {
-    this.setState(({ notifications }) => ({
-      notifications: notifications.filter(({ id }) => id !== notificationId)
-    }))
+    this.props.hideNotification(notificationId)
```
Partysun: instead of notificationId, we can use a simple id @devanshu-santiment

src/components/NotificationStack/index.js
devanshu-santiment: @Partysun the index here ("i") is used to calculate the position on the y axis for each element.
Should I find an alternate way of just renaming it to `index` will work ?

src/components/NotificationStack/index.js
Partysun: in this function

src/components/NotificationStack/index.js
```diff
@@ -1,10 +1,12 @@
 import React, { Component } from 'react'
 import { connect } from 'react-redux'
+import { bindActionCreators } from 'redux'
 import { Notification as NotificationItem } from '@santiment-network/ui'
 import { CSSTransition, TransitionGroup } from 'react-transition-group'
+import * as Actions from '../../actions/rootActions'
```
Partysun: We used "actions" instead of "Actions" @devanshu-santiment

src/components/NotificationStack/index.js
Partysun: @devanshu-santiment rename it for faster readability.

PR #147, opened 2019-04-01
feature/new-trending-words-page

DmitriVanGuard:
### Summary
First iteration of the new `Trends` lab page.
Getting change of the `Trending score` using [this API](https://api-stage.santiment.net/graphiql?query=query%20wordTrendScore(%24word%3A%20String!%2C%20%24from%3A%20DateTime!%2C%20%24to%3A%20DateTime!)%20%7B%0A%20%20wordTrendScore(word%3A%20%24word%2C%20source%3A%20ALL%2C%20from%3A%20%24from%2C%20to%3A%20%24to)%20%7B%0A%20%20%20%20datetime%0A%20%20%20%20score%0A%20%20%7D%0A%7D&variables=%7B%0A%20%20%22word%22%3A%20%22april%22%2C%0A%20%20%22from%22%3A%20%222019-03-30T07%3A39%3A48.873Z%22%2C%0A%20%20%22to%22%3A%20%222019-03-31T23%3A39%3A48.873Z%22%0A%7D).
Getting change of the `Social volume` using [this API](https://api-stage.santiment.net/graphiql?query=query%20topicSearch(%24searchText%3A%20String!%2C%20%24from%3A%20DateTime!%2C%20%24to%3A%20DateTime!)%20%7B%0A%20%20topicSearch(searchText%3A%20%24searchText%2C%20source%3A%20TELEGRAM%2C%20from%3A%20%24from%2C%20to%3A%20%24to%2C%20interval%3A%20%221d%22)%20%7B%0A%20%20%20%20chartData%20%7B%0A%20%20%20%20%20%20datetime%0A%20%20%20%20%20%20mentionsCount%0A%20%20%20%20%7D%0A%20%20%20%20%0A%20%20%7D%0A%7D&variables=%7B%0A%20%20%22from%22%3A%20%222019-03-28T21%3A00%3A00.000Z%22%2C%0A%22searchText%22%3A%20%22rcn%22%2C%0A%22to%22%3A%20%222019-03-31T21%3A00%3A00.000Z%22%0A%7D) and combining results form different sources
#### Changes
- New view for the `/labs/trends`
- Fixed apollo-cache issues for the `WordCloud`, `SocialVolume` and trends word epic.
- Removed artificial delay in loading trend words. It's not necessary after the first fetch, because the data is cached.

### Screenshots
image
image


src/components/SocialVolumeWidget/socialVolumeEpic.js
```diff
@@ -20,15 +19,19 @@ export const fetchSocialVolumeEpic = (action$, store, { client }) =>
       const isFetchingTotalSocialVolume =
         trendWord === actions.TOTAL_SOCIALVOLUME_SECRET
 
+      const dateTo = new Date()
```
Partysun: Move to util module for time. @DmitriVanGuard

src/components/ValueChange/ValueChange.js
```diff
@@ -0,0 +1,30 @@
...
+const Change = {
+  true: ['jungle-green', 'triangle-up'],
+  false: ['persimmon', 'triangle-down']
+}
+
+const notChanged = ['texas-rose', 'lock-small']
+
+const ValueChange = ({ oldValue, newValue, render }) => {
+  const amp = newValue - oldValue
+  const [accent, triangle] = amp !== 0 ? Change[amp > 0] : notChanged
```
Partysun: Explicit maybe better from fast readability than 2 vars and game with types.
I mean {
positive:
negative:
notChanged:
}

src/components/WordCloud/fetchWordContextEpic.js
```diff
@@ -19,10 +18,12 @@ export const preloadWordContextEpic = (action$, store, { client }) =>
         return Observable.of()
       }
 
-      const dateTo = moment().toISOString()
-      const dateFrom = moment()
-        .subtract(3, 'd')
-        .toISOString()
+      const dateTo = new Date()
```
Partysun: same

src/components/WordCloud/fetchWordContextEpic.js
```diff
@@ -75,15 +77,19 @@ export const fetchWordContextEpic = (action$, store, { client }) =>
         })
       }
 
+      const dateTo = new Date()
```
Partysun: same

src/epics/rootEpics.js
```diff
@@ -89,7 +93,7 @@ export default combineEpics(
   fetchAllTickersSlugs,
   // WordCloud
```
Partysun: comments

src/pages/Trends/TotalSocialVolume.js
```diff
@@ -0,0 +1,102 @@
...
+            datetime: longestTSData.datetime
+          }
+        }
+      }),
+      loading
+    }
+  },
+  options: () => {
+    const toDate = new Date()
+    toDate.setHours(24, 0, 0, 0)
```
Partysun: and again

src/pages/Trends/changesEpic.js
```diff
@@ -0,0 +1,136 @@
...
+const defaultMentionsCount = {
+  mentionsCount: 0
+}
+
+export const wordTrendSocialVolumeEpic = (action$, store, { client }) =>
+  action$.ofType('[trends] HYPED_FETCH_SUCCESS').mergeMap(({ payload }) => {
+    // HACK(vanguard): wordTrendScore from/to does not work correctly
+    // Can't fetch only needed time period, should fetch for all day
+    const fromDate = new Date()
+    fromDate.setHours(0, 0, 0, 0)
```
Partysun: one more

src/pages/Trends/changesEpic.js
```diff
@@ -0,0 +1,136 @@
...
+        })
+      })
+      .catch(handleErrorAndTriggerAction('changes failed'))
+  })
+
+export const wordTrendScoreEpic = (action$, store, { client }) =>
+  action$.ofType('[trends] HYPED_FETCH_SUCCESS').mergeMap(({ payload }) => {
+    // HACK(vanguard): wordTrendScore from/to does not work correctly
+    // Can't fetch only needed time period, should fetch for all day
+    const fromDate = new Date()
```
Partysun: and one more

src/pages/Trends/changesEpic.js
```diff
@@ -0,0 +1,136 @@
...
+        }
+      })
+    })
+
+    return Observable.forkJoin(...words)
+      .flatMap(result => {
+        return Observable.of({
+          type: TREND_WORD_SCORE_CHANGE_FULFILLED,
+          payload: result.reduce((acc, value) => {
+            Object.assign(acc, value)
```
Partysun: @DmitriVanGuard Why not spread operator. Node.js team said: "The object rest/spread operators are both syntactically neat and offer performance benefits over Object.assign()"
Sure it's a browser, but I don't think you win a lot here. If you don't have much better performance, better to use Spread. And for safety, better to use Object.assign({}, acc, value)

src/pages/Trends/TotalSocialVolume.js
DmitriVanGuard: I'm really against this. Current code and other examples contain 0 complicated logic, neither it requires any documentation (especially when the standardized `Date` interface is already well documented).

**Single instructions** should not be placed in the utils method. Here is the reason:
It just adds more weight to dependency management. Firstly, the developer needs to remember how the util function is named, then think of where it's located relative to the current directory and then write an `import` and finally use it. And all that for just using the simple native instruction that every JS developer should be familiar with. 

In the end, this *utils* pattern just add more weight to the app bundle and requires to handle all these indirections

src/pages/Trends/changesEpic.js
DmitriVanGuard: 1. It's an array loop where the performance matter;
2. No need for `immutability` (Object.assign({}, acc, value)), because the `acc` is used only in `reduce` context;
3. `Object.assign()` is standartized, doesn't require pollyfils and without extra allocations is significantly faster.
Benchmark: https://jsperf.com/object-assign-vs-spread
image

src/components/ValueChange/ValueChange.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
```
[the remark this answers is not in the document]
DmitriVanGuard: @Partysun `ValueChange` was created now as a separate component for feature use. It has a much more versatile interface, which is not compatible with current `PercentChanges`.  `PercentChanges` is used in the components with an old design which are targets of large refactoring. Also, it's much better to keep PR changes related to the only specific idea, which in our case is 'trends'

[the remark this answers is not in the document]
Partysun: @DmitriVanGuard Upgrade PercentChange, remove ValueChange. It's duplicated by logic. We kill 2 problems. First, remove font-awesome usage in PercentChange. You can add old/new value as options. It's easiest and bad practice to do any new staff from scratch.
 
Keep PR changes related to one specific idea means not to make new classes for any your idea. If you need such thing as ValueChange, better to make separate PR. It's one of the reasons why we have big PRs. When you are doing a big feature, you make modular things in this pr.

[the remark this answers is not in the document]
DmitriVanGuard: @Partysun Please, see what the `PercentChange`. It has completely different logic which is also implied by its name. `PercentChange` always appends a percent sign a the end of the value and have weir interface. If we make percent sign optional, then suprise-suprise its becomes `ValueChange` 😄. `ValueChange` was created because `PercentChange` could not be used in the design implementation. 

Speaking of bad practices: Adding props whenever we want to **IS** a bad practice which inflates the logical borders of the components. As I mentioned before, `PercentChange` could be composed by `ValueChange` which is **recommended** pattern by the [core React team](https://reactjs.org/docs/composition-vs-inheritance.html).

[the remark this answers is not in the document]
Partysun: @DmitriVanGuard About upgrade PercentChange I mean to make anything inside this class and optional rename this module. If we make one more basic class than we have to refactor old class and use composition but make such a new class without refactoring old, you produce 2 branches. Make separate PR and refactor all usages of old class with new changes/basic class. 
Add story for this class.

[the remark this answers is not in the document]
DmitriVanGuard: Sure, that was exactly one of my points! The old component will be refactored with an update of all dependencies in separate PR 👍

src/utils/dates.js
```diff
@@ -0,0 +1,20 @@
+const DateFormat = {
+  m: ['getMonth', 'setMonth'],
+  d: ['getDate', 'setDate']
+}
+
+export const getTimeIntervalFromToday = (amount, dateFormat) => {
+  const fromDate = new Date()
```
Partysun: @DmitriVanGuard Maybe in this function better to use 'from' and 'to'. As return object has params.
What do you think?

DmitriVanGuard:
Hm, I guess function like this will be used only in cases with general time
intervals (like in cases in current PR) which does not have follow up (or
even prior) modification of the date. I think it’s okay to create from/to
here in order to remove more code repetition

вт, 2 апр. 2019 г. в 22:17, Yura <notifications@github.com>:

src/utils/dates.js
DmitriVanGuard: @Partysun Sure 👌, updated variables names to `from` and `to`

src/components/PercentChanges.js
```diff
@@ -1,35 +1,16 @@
...
-      {normalizedChanges < 0
-        ? normalizedChanges.toString().split('-')[1]
-        : normalizedChanges}
-      %
-    </div>
-  )
-}
+const render = change => `${parseFloat(change).toFixed(2)} %`
+
+const PercentChanges = ({ className, changes }) => (
```
Partysun: 👍 This is what I want ;)

PR #153, opened 2019-04-03
feature/dashboard-page

DmitriVanGuard:
### Summary
The first iteration of the `dashboard` page. 
Currently on this page are available only `Last trends` and `Featured insights`.

#### Changes
- The default page is now `/dashboard`;
- Fallbach is now `/dashboard`.

### Screenshots
image


src/components/Insight/ProfileInfo.js
```diff
@@ -9,20 +9,25 @@ const ProfileInfo = ({
...
-        {picUrl ? (
-          <img src={picUrl} alt='Profile Pic' />
-        ) : (
-          <Icon type='profile-round' />
-        )}
-        {networkStatus && (
-          <div className={cx(styles.onlineIndicator, styles[networkStatus])} />
-        )}
-      </div>
+      {!noPic && (
```
Partysun: @DmitriVanGuard Logic here is "no" + "no" + Pic && (make smth). Condition starts with negative is less readable.
Possible better make withPic. Default value is true. And condition withPic &&.

Also, we can move all code after && () to a variable before return. Or we can remove/reduce "return".

src/pages/Trends/TrendsPage.module.scss
```diff
@@ -46,3 +39,77 @@
...
+
+  &__right {
+    flex: 1;
+    display: flex;
+    flex-direction: column;
+  }
+}
+
+.subtitle {
+  font-family: Rubik;
```
Partysun: @DmitriVanGuard Do we need to have this line?

DmitriVanGuard: Unfortunately - yes. There are global styles that should be removed in future css refactoring PR (it's a breaking change) 
image

PR #154, opened 2019-04-04
feature/connected-trending-words

DmitriVanGuard:
### Summary
Highlighting connected trending words by hovering on the corresponding action button.

### Screenshots
image

### Algorithm:
**NOTE**: Here is described only core of the algorithm. In the code there is a place for optimization and special cases.

**Known data**
- List of `trending word`s;
- List of projects with its `ticker`, `slug` and `name` in it;
- Every insight has in its tags project `ticker` or a custom word.

**Steps**
1. Fetching insights by `*-trending-words` tag for the current and last 2 days;
2. Building the `graph of tags`; 
3. Connecting `trending word`s to the project's `ticker` by searching in the array of all project's and comparing it to the `ticker` or `slug`, or by checking if it's included in the project's `name`. 
4. Searching in the `graph of tags` for a tag similar to the:
    -  `trending word`;
    - connected `trending word` `ticker`;
5.  Mapping every connection of the found `tag` to the `trending word` (or `trending word` ticker)

**Steps explanation**
1. If today's date is April 4th, then fetching insights with tags `04-04-trending-words`, `03-04-trending-words`, `02-04-trending-words`;
2. E.g. insight has these tags: `BTC`, `ETH`, `LTC`. The graph will look like this: 
```js
const graph = {
  BTC: ["ETH", "LTC"],
  ETH: ["BTC", "LTC"],
  LTC: ["ETH", "BTC"],
}
```
3. E.g. trending word like `bitcoin` is the same as the ticker `BTC`;
4. Checking if the `graph["BTC"]` exist;
5. E.g. trending words are `bitcoin`, `doge`, `litecoin`. On the step `3` they are mapped to corresponding tickers `BTC`, `DOGE`, `LTC`. Now, substitute for the matched `tags` and `trending word`s is taken place. The result will be:
```
const result = {
  bitcoin: ["litecoin"],
  litecoin: ["bitcoin"]
}
```
`ETH` is not present in the result, because there were no `trending word` match for it


src/pages/Trends/connectedWordsEpic.js
```diff
@@ -0,0 +1,234 @@
...
+
+const tickerCheckFn = (target, { ticker }) => target === ticker
+const tickerMoveFn = (target, { ticker }) => target < ticker
+const slugCheckFn = (target, { slug }) => target === slug
+const slugMoveFn = (target, { slug }) => target < slug
+
+let projectsSortedByTicker
+let projectsSortedBySlug
+
+const basicSort = (a, b) => {
```
Partysun: https://github.com/santiment/app/blob/master/src/utils/sortMethods.js
@DmitriVanGuard

src/pages/Trends/connectedWordsEpic.js
```diff
@@ -0,0 +1,234 @@
...
+} from '../../components/Trends/actions'
+import { ALL_INSIGHTS_BY_TAG_QUERY } from '../../components/Insight/insightsGQL'
+import { binarySearch } from './utils'
+
+const oneDayTimestamp = 1000 * 60 * 60 * 24
+
+const getInsightTrendTagByDate = date =>
+  `${date.getDate()}-${date.getMonth()}-${date.getFullYear()}-trending-words`
+
+const tickerCheckFn = (target, { ticker }) => target === ticker
```
Partysun: Why did you use Fn in names? @DmitriVanGuard

src/pages/Trends/utils.js
```diff
@@ -0,0 +1,36 @@
+const defaultMoveFn = (target, value) => target < value
+const defaultCheckFn = (target, value) => target === value
+
+export const binarySearch = ({
```
Partysun: @DmitriVanGuard I think I saw before bs function in our codebase

src/pages/Trends/connectedWordsEpic.js
```diff
@@ -0,0 +1,234 @@
...
+
+export const connectedWordsOptimizationEpic = action$ =>
+  action$
+    .ofType(TRENDS_HYPED_FETCH_TICKERS_SLUGS_SUCCESS)
+    .take(1)
+    .switchMap(({ payload: { allAssets } }) => {
+      const { length } = allAssets
+      const tempAssets = allAssets.slice()
+
+      for (let i = 0; i < length; i++) {
```
Partysun: why not map? @DmitriVanGuard

src/pages/Trends/connectedWordsEpic.js
DmitriVanGuard: Really was not thinking about it when created. Maybe better to call it with `Clb` at the end?

src/pages/Trends/connectedWordsEpic.js
DmitriVanGuard: Slower for an array of 1000+ elements with an allocation memory for a new object on every iteration

src/pages/Trends/utils.js
DmitriVanGuard: @Partysun I have [found it](https://github.com/santiment/app/blob/c017a705077988430e006875cbfad5d35f60b952/src/utils/utils.js#L87). It's currently optimized for finding correct history price (basically for datetimes) and I'm a little bit afraid to modify it now using new `binarySearch` function (which is general and can be used in whatever case). I will do it in a separate PR after merging this one

PR #158, opened 2019-04-05
Add custom scroll bar for webkit

Partysun:
Clean ProjectIcon

### DIFF
--- a/src/index.scss
+++ b/src/index.scss
@@ -111,6 +111,24 @@ hr {
   z-index: 100;
 }
 
+// Custom Scrollbar
+::-webkit-scrollbar {
+  width: 4px;
+  border-radius: 4px;
+}
+
+::-webkit-scrollbar-track {
+  background: var(--porcelain);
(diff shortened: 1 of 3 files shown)


src/index.scss
```diff
@@ -111,6 +111,24 @@ hr {
   z-index: 100;
 }
 
+// Custom Scrollbar
+::-webkit-scrollbar {
```
DmitriVanGuard: Scroll bar styles are already specified in the `san-ui` [`main.scss`](https://github.com/santiment/san-ui/blob/47ced098a0d6874e98693fa9f6cf5185076d4008/src/main.scss#L113), which is imported to the `src/index.scss` [here](https://github.com/santiment/app/blob/6cfc8aa56ccdadc75c9201f70fa5000341b9dec4/src/index.scss#L3)

Partysun: @DmitriVanGuard nice. didn't notice

PR #160, opened 2019-04-07
feature/time-from-now-util

DmitriVanGuard:
#### Summary
Santiment alternative to the [`moment().fromNow()`](https://momentjs.com/docs/#/displaying/fromnow/) method.

#### Benchmarks
[Benchmark](https://jsperf.com/time-from-now) of the current implementation. On average, this implementation is `18x` faster than `moment` and `2x` faster than `date-fns`
image


Partysun:
Benchmark look good, but sure for us it's less important than reducing size of bundle) But great job.
Add please JSDOC string for this util method.

PR #162, opened 2019-04-08
Feature/mobile navbar

Partysun:
image


src/components/MobileNavbar/MobileNavbarAction.js
```diff
@@ -0,0 +1,45 @@
...
+      )}
+      isActive={isActive}
+      onClick={handleOnClick}
+    >
+      <Icon
+        type={iconType}
+        className={classes.icon}
+        fill={isActive ? 'var(--jungle-green)' : '#000'}
+      />
+      <div className={cx(classes.label, { [classes.isActive]: isActive })}>
```
DmitriVanGuard: `cx(classes.label, { [classes.isActive]: isActive })` could be much simplier, without passing an extra object, since there are no more conditional classes:`cx(classes.label, isActive && classes.isActive })`

src/components/MobileNavbar/MobileNavbar.js
```diff
@@ -0,0 +1,183 @@
...
+              linkTo={linkTo}
+              iconType={iconType}
+              onClick={handleNavigation}
+              label={label}
+            />
+          )
+        })}
+        <MobileNavbarAction
+          onClick={toggleMenu}
+          iconType={isOpened ? 'close' : 'hamburger'}
```
DmitriVanGuard: `iconType` could be calculated inside `MobileNavbarAction` via `isActive`. @Partysun What do you think?

Partysun: @DmitriVanGuard yes, but than MobileNavbarAction will be more specific.
We can possibly to move it to san-ui later.

src/components/MobileNavbar/MobileNavbar.module.scss
```diff
@@ -0,0 +1,72 @@
...
+      justify-content: center;
+      margin: 24px 16px;
+    }
+
+    .navigationList {
+      display: flex;
+      flex-direction: column;
+    }
+
+    .navigationList a {
```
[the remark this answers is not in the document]
Partysun: @DmitriVanGuard What is the benefit? (Speed?)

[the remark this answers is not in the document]
DmitriVanGuard: Yes, same as `[1]`. As first step browsers get all the `a` tags and then applies a filter for all selectors of the `descendant selector` step by step and that on top of all other classes and selectors

PR #172, opened 2019-04-09
Add MobileHeader

Partysun:
Screenshot 2019-04-10 at 01 13 17

### DIFF
--- a/src/components/MobileHeader/MobileHeader.js
+++ b/src/components/MobileHeader/MobileHeader.js
@@ -1,0 +1,46 @@
+import React, { Fragment } from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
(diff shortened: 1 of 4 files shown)


src/components/MobileHeader/MobileHeader.js
```diff
@@ -0,0 +1,46 @@
...
+import cx from 'classnames'
+import { Link } from 'react-router-dom'
+import { Button, Icon } from '@santiment-network/ui'
+import styles from './MobileHeader.module.scss'
+
+const MobileHeader = ({
+  title,
+  backRoute,
+  rightActions,
+  classes = {
```
devanshu-santiment: @Partysun would be better if you define a const (may be `defaultClasses` or some better name) at the top and export it. Would be possible to extend from parent component without providing all the styles

PR #174, opened 2019-04-09
Mobile Insights page

Partysun:
image

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -46,12 +46,6 @@ const LoadableDetailedPage = Loadable({
   loading: () => <PageLoader />
 })
 
-const LoadableInsights = Loadable({
-  loader: () => import('./pages/Insights/'),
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -193,9 +194,17 @@ export const App = ({
...
           path='/insights'
-          render={props => (
-            <LoadableInsights {...props} isLoggedIn={isLoggedIn} />
-          )}
+          component={Loadable({
+            loader: () => import('./pages/Insights/'),
+            render: (loaded, props) => (
+              <loaded.default
+                isDesktop={isDesktop}
+                isLoggedIn={isLoggedIn}
```
DmitriVanGuard: `isDesktop` could be passed here `<LoadableInsights {...props} isLoggedIn={isLoggedIn} />` as already done with the `isLoggedIn` prop + we do it like this in other places

PR #181, opened 2019-04-10
Feature/dashobard watchlist

devanshu-santiment:
screencapture-localhost-3000-dashboard-2019-04-10-23_52_27


src/pages/assets/AssetsOverviewPage.js
```diff
@@ -72,7 +72,9 @@ const AssetsOverview = props => (
         )
       )}
     </div>
-    <FeaturedWatchlist />
+    <div className={styles.flexRow}>
```
Partysun: @devanshu-santiment I think row is better. But it's ok. Let's merge

PR #219, opened 2019-04-13
fix/digest-form

DmitriVanGuard:
#### Summary
Fix for the non-approved GDPR users


src/epics/handleEmailLogin.js
```diff
@@ -61,43 +61,59 @@ const NEWSLETTER_SUBSCRIPTION_MUTATION = gql`
...
+export const digestSubscriptionEpic = (action$, store, { client }) =>
+  action$
+    .ofType(actions.USER_LOGIN_SUCCESS)
+    .take(1)
+    .mergeMap(({ user: { privacyPolicyAccepted } }) =>
+      (privacyPolicyAccepted
+        ? Observable.of(true)
+        : action$.ofType(actions.USER_SETTING_GDPR)
+      )
+        .delayWhen(() => Observable.timer(2000))
```
Partysun: @DmitriVanGuard Why 2000?

DmitriVanGuard: @Partysun If the request is sent immediately, the server might not be able to update internal data and will respond with `unauthorized` or `privacy policy is not accepted`. Happened several times while testing

PR #225, opened 2019-04-15
enhance/connected-trends

DmitriVanGuard:
#### Summary
- Fixed bug with mapping insights to a new found trend ticker;
- Highlighting connected trends in neighbor trend tables;
- Connected trends number-label now is relative to the visible trends;
- Fixed the inconsistent behavior of auto-selected tags from trends when writing new insight;
- Keeping `connectedTrends` up-to-date.

#### Screenshots
image

### DIFF
--- a/src/components/Trends/TrendsTable/TrendsTables.js
+++ b/src/components/Trends/TrendsTable/TrendsTables.js
@@ -63,11 +63,12 @@ class TrendsTables extends PureComponent {
       connectedTrends
     } = this.props
 
+    const { length } = trends
+
     return (
       <div className={styles.tables}>
         {trends.length > 1 &&
-          trends.slice(0, -1).map(trend => {
-            const { datetime } = trend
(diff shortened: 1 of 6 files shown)


Partysun:
@DmitriVanGuard Change "Trending Score" to "Hyped Score"

src/components/Trends/TrendsTable/TrendsTables.js
```diff
@@ -68,17 +97,22 @@ class TrendsTables extends PureComponent {
...
 }
 
 const mapStateToProps = ({
-  hypedTrends: { selectedTrends },
+  hypedTrends: { selectedTrends, connectedTrends },
   user: { token }
 }) => ({
   selectedTrends,
+  connectedTrends,
   isLoggedIn: !!token
```
Partysun: @DmitriVanGuard What do you think. Let's use for loggedin one selector function.
Cause if we will change logic of this. There are a lot of changes, we will do.

PR #226, opened 2019-04-15
Fix/insights visual backtest

Partysun:
This is PR was made for change updatedAt to publishedAt for proper calculation of VisualBackteset Chart. But I've noticed a lot of the same code in several parts of our codebase.
The idea is to make one source of truth for Insights related queries. Maybe it's not the best place, but we need to keep our tree flexible.

But during this structure changes spec for connectedWords was broken. Can @DmitriVanGuard check this PR and this spec.


Partysun:
Feature works image

Partysun:
@DmitriVanGuard Can you add required filed in the queries, which need to have this vars.
Yes about logging. This pr is a proposal, not was finished.

PR #227, opened 2019-04-15
Enhance/featured insights styles

haritonasty:
scrollbar on hover. There is a problem with bias content when scrollbar appear. For webkit- I used to non-standard `overflow: overlay`. In firefox when you hover on the block - cards decrease their widths for some px (depends on scrollbar' width)

image

### DIFF
--- a/src/components/Insight/InsightCardSmall.js
+++ b/src/components/Insight/InsightCardSmall.js
@@ -1,9 +1,9 @@
 import React from 'react'
 import { Link } from 'react-router-dom'
 import LikeBtn from '../Like/LikeBtn'
+import MultilineText from '../MultilineText/MultilineText'
 import { getSEOLinkFromIdAndTitle } from '../../pages/Insights/utils'
 import styles from './InsightCardSmall.module.scss'
-import MultilineText from '../MultilineText/MultilineText'
 
 const InsightCard = ({
   className = '',
--- a/src/components/Insight/InsightCardSmall.js
+++ b/src/components/Insight/InsightCardSmall.js
@@ -1,6 +1,5 @@
 import React from 'react'
 import { Link } from 'react-router-dom'
-import moment from 'moment'
 import LikeBtn from '../Like/LikeBtn'
 import { getSEOLinkFromIdAndTitle } from '../../pages/Insights/utils'
 import styles from './InsightCardSmall.module.scss'
@@ -9,7 +8,7 @@ const InsightCard = ({
   className = '',
   id,
   title,
-  createdAt,
+  user,
   votedAt,
   votes: { totalVotes }
 }) => {
@@ -22,8 +21,8 @@ const InsightCard = ({
         {title}
       </Link>
       <div className={styles.meta}>
-        <div className={styles.date}>
-          {moment(createdAt).format('MMM D, YYYY')}
+        <div className={styles.username} title={user.username}>
+          {user.username}
         </div>
         <LikeBtn small grey liked={!!votedAt} likesNumber={totalVotes} />
       </div>
--- a/src/components/Insight/InsightCardSmall.js
+++ b/src/components/Insight/InsightCardSmall.js
@@ -3,6 +3,7 @@ import { Link } from 'react-router-dom'
 import LikeBtn from '../Like/LikeBtn'
 import { getSEOLinkFromIdAndTitle } from '../../pages/Insights/utils'
 import styles from './InsightCardSmall.module.scss'
+import MultilineText from '../MultilineText/MultilineText'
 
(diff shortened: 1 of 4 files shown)


src/components/Insight/InsightCardSmall.js
```diff
@@ -22,8 +21,8 @@ const InsightCard = ({
         {title}
       </Link>
       <div className={styles.meta}>
-        <div className={styles.date}>
-          {moment(createdAt).format('MMM D, YYYY')}
+        <div className={styles.username} title={user.username}>
+          {user.username}
```
DmitriVanGuard: Should be a link to the insights filtered by the author, e.g: `/insights/users/7`

src/components/Insight/InsightCardSmall.module.scss
```diff
@@ -1,12 +1,18 @@
...
   color: var(--mirage);
   max-width: 100%;
-  overflow-x: hidden;
+  margin-bottom: 6px;
+  overflow: hidden;
   text-overflow: ellipsis;
-  white-space: nowrap;
+  word-wrap: break-word;
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
```
DmitriVanGuard: Custom and cross-platform component for multiline text truncating -> [`MultilineText`](https://github.com/santiment/app/blob/master/src/components/MultilineText/MultilineText.js)

Partysun:
@haritonasty  @DmitriVanGuard I think we will change a style of this featured block from overflow: hidden to scroll onMouseEnter/Leave events. Or is it not necessary for full browser support?

haritonasty:
> @haritonasty @DmitriVanGuard I think we will change a style of this featured block from overflow: hidden to scroll onMouseEnter/Leave events. Or is it not necessary for full browser support?

haritonasty:
@Partysun 
I think, it can't solve problem. Now this is working everywhere, but in non-webkit - small bias, because scrollbar has width.

src/components/Insight/InsightCardSmall.module.scss
haritonasty: done

src/components/Insight/InsightCardSmall.js
haritonasty: done

src/components/Insight/InsightCardSmall.module.scss
```diff
@@ -3,10 +3,9 @@
   font-size: 16px;
   line-height: 24px;
   color: var(--mirage);
-  max-width: 100%;
-  overflow-x: hidden;
-  text-overflow: ellipsis;
-  white-space: nowrap;
+  max-width: 400px;
+  margin-bottom: 6px;
+  word-wrap: break-word;
```
DmitriVanGuard: `word-wrap` is regulated by the `MultilineText`. `max-width` in such a scenario is not necessary

src/components/Insight/InsightCardSmall.module.scss
```diff
@@ -15,11 +14,16 @@
 
 .meta {
   display: flex;
+  justify-content: space-between;
 }
 
-.date {
-  width: 100px;
+.username {
+  max-width: 200px;
```
DmitriVanGuard: `max-width` is not necessary because of the `flex` parent

src/components/Insight/InsightCardSmall.js
```diff
@@ -1,15 +1,15 @@
 import React from 'react'
 import { Link } from 'react-router-dom'
-import moment from 'moment'
 import LikeBtn from '../Like/LikeBtn'
 import { getSEOLinkFromIdAndTitle } from '../../pages/Insights/utils'
 import styles from './InsightCardSmall.module.scss'
```
DmitriVanGuard: We have a convention to put the imported `styles` at the bottom of all imports

src/components/Insight/InsightCardSmall.module.scss
haritonasty: @DmitriVanGuard about max-width: in this widget - yes, but in "Emerging trends"  when hover on last icon in row - it causes problems without max-width

Partysun:
I see the difference between this and master design
image

haritonasty:
> Read my comment above

@Partysun hmm...if I make it wider, I'll break grid. left block will be too small

Partysun:
I am sorry, but :) - border-bottom-left-radius @haritonasty 
image

haritonasty:
@Partysun you are so big-eyed))

Partysun:
:) @haritonasty One more we have updated version in Figma for desktop of this block)
image

PR #229, opened 2019-04-16
enhance/trends-table

DmitriVanGuard:
#### Summary
- Explanation tooltips for the action buttons;
- Action buttons now trigger tooltip by `click`;
- `Hyped score` renamed to the `Hype score`
- By clicking on the table row, the user will be lead to the `/labs/trends/explore/:trend` page;

#### Screenshots
image

### DIFF
--- a/src/components/Trends/TrendsTable/TrendsTable.js
+++ b/src/components/Trends/TrendsTable/TrendsTable.js
@@ -11,7 +11,7 @@ import {
   Panel,
   Icon,
   Tooltip,
-  Button
+  Button,
 } from '@santiment-network/ui'
 import { store } from '../../../index'
 import ValueChange from '../../../components/ValueChange/ValueChange'
@@ -24,20 +24,20 @@ const columns = [
     Header: '#',
     accessor: 'index',
     width: 35,
-    headerClassName: styles.headerIndex
+    headerClassName: styles.headerIndex,
   },
   {
     Header: 'Word',
-    accessor: 'word'
+    accessor: 'word',
   },
   {
-    Header: 'Hyped score',
-    accessor: 'score'
+    Header: 'Hype score',
+    accessor: 'score',
   },
   {
     Header: 'Social volume',
-    accessor: 'volume'
-  }
+    accessor: 'volume',
+  },
 ]
 
 const NumberCircle = ({ className, ...props }) => (
@@ -70,7 +70,7 @@ const getTrGroupProps = (_, rowInfo) => {
         node = node.parentNode
       }
       store.dispatch(push(`/labs/trends/explore/${rowInfo.original.rawWord}`))
-    }
+    },
   }
 }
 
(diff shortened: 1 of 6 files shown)


src/components/Trends/TrendsTable/TrendsTable.js
```diff
@@ -3,14 +3,17 @@ import Table from 'react-table'
...
+import { push } from 'react-router-redux'
 import {
   Label,
   Checkbox,
   PanelWithHeader,
   Panel,
   Icon,
-  Tooltip
+  Tooltip,
+  Button,
```
Partysun: @DmitriVanGuard we are not python. What is the new trend to use "," at the end?

src/components/Trends/TrendsTable/TrendsTable.js
```diff
@@ -21,116 +24,134 @@ const columns = [
...
+      while (node && node !== currentTarget) {
+        if (
+          node.classList.contains(styles.action) ||
+          node.classList.contains(styles.checkbox)
+        ) {
+          return
+        }
+        node = node.parentNode
+      }
+      store.dispatch(push(`/labs/trends/explore/${rowInfo.original.rawWord}`))
```
Partysun: @DmitriVanGuard  We loose here "a" ability to open this in new tab. And spiders lose these links.

src/components/Trends/TrendsTable/TrendsTable.js
```diff
@@ -21,116 +24,134 @@ const columns = [
...
+  },
 ]
 
-const NumberCircle = props => (
-  <div {...props} className={styles.insights__number} />
+const NumberCircle = ({ className, ...props }) => (
+  <div {...props} className={cx(className, styles.insights__number)} />
+)
+
+const Explanation = ({ text, children, ...props }) => (
```
Partysun: Is it possible common usage for many tooltips on our UI? @DmitriVanGuard

src/components/Trends/TrendsTable/TrendsTable.js
DmitriVanGuard: Prettier

src/components/Trends/TrendsTable/TrendsTable.js
DmitriVanGuard: No, we don't and [here is why](https://github.com/santiment/app/blob/2aa1a1174a2cd06c5dd7e0add1b388c89a323865/src/components/Trends/TrendsTable/TrendsTable.js#L246)

src/components/Trends/TrendsTable/TrendsTable.js
Partysun: Oh, yes. Don't notify it before :)

src/components/Trends/TrendsTable/TrendsTable.js
DmitriVanGuard: Actually 2 or 3 places. So yeah, better to move it to the separate file 👍

PR #230, opened 2019-04-16
Enhance/featured insights to feed

haritonasty:
mobile - https://www.figma.com/file/RPLINi2nrD30hQjcANws3CxH/Sanbase?node-id=4911%3A7
tablet - https://www.figma.com/file/RPLINi2nrD30hQjcANws3CxH/Sanbase?node-id=5300%3A0
desktop - https://www.figma.com/file/RPLINi2nrD30hQjcANws3CxH/Sanbase?node-id=1852%3A35

mobile result: 
image

tablet result:
image

desktop result: 
image

I used two layouts, because on mobile & tabled it located between feeds, on desktop - in sidebar. And I could rewrote to one layout, but sidebar in desktop will be with absolute positioning - this way isn't so flexible, as I think.


src/components/Feed/Feed.js
```diff
@@ -1,13 +1,32 @@
...
+  </section>
+)
+
 const Feed = ({ component: El, data, dateKey }) => {
   let lastDateKey
   return data.map((item, index) => {
     const id = item.id || index
     const date = moment(item[dateKey]).format('MMM D')
     const isNotSameAsLastDate = date !== lastDateKey
+    const showAfterElNumber = 2
```
DmitriVanGuard: This variable should be defined statically and named as the constant. Currently, it's created on every `Feed` rerender

src/components/Feed/Feed.js
```diff
@@ -1,13 +1,32 @@
 import React, { Fragment } from 'react'
 import moment from 'moment'
+import InsightsFeatured from '../Insight/InsightsFeatured'
 import styles from './Feed.module.scss'
 
+const FeaturedInsightsBlock = () => (
```
DmitriVanGuard: Should be in a separate file along with styles.

src/components/Feed/Feed.js
```diff
@@ -17,6 +36,7 @@ const Feed = ({ component: El, data, dateKey }) => {
       <Fragment key={id}>
         {isNotSameAsLastDate && <h4 className={styles.date}>{date}</h4>}
         <El className={styles.signal} {...item} />
+        {index === showAfterElNumber && <FeaturedInsightsBlock />}
```
DmitriVanGuard: The main flaw in the current logic is that `Featured insights` view change between desktop/mobile devices happens via css. That's a lot of additional calls + memory garbage that React should handle and reconcile => big chunk of unnecessary weight for mobile devices.

**How we can solve it?**
There should be only one `InsightsFeaturedBlock` in two different places (at the side and in the middle of the feed), that should be conditionally rendered via `isDesktop` prop

src/pages/Insights/InsightsPage.module.scss
```diff
@@ -25,15 +25,16 @@
 }
 
 .mainSection {
-  padding: 0 16px;
+  padding: 0;
```
DmitriVanGuard: @haritonasty  What was the reason for removing `16px`?

src/components/Feed/Feed.module.scss
```diff
@@ -1,17 +1,118 @@
...
+
+  color: var(--grey-dark-2);
+  margin: 40px 16px 16px 16px;
 }
 
 .signal {
-    margin-bottom: 24px;
+  margin-bottom: 24px;
+  margin-right: 16px;
+  margin-left: 16px;
```
DmitriVanGuard: Line 13-15 should be written as a shorthand

src/components/Feed/Feed.module.scss
```diff
@@ -1,17 +1,118 @@
...
+@import "../../mixins.scss";
 
 .date {
-    @include text-normal;
-    color: var(--grey-dark-2);
-    margin: 40px 0 16px;
+  @include text-normal;
+
+  color: var(--grey-dark-2);
+  margin: 40px 16px 16px 16px;
```
DmitriVanGuard: Shorthand `40px 16px 16px;`

src/pages/Insights/InsightsPage.module.scss
```diff
@@ -25,15 +25,16 @@
...
 }
 
 .newSignal {
-  width: 100%;
   margin-bottom: 12px;
   display: flex;
   align-items: center;
   justify-content: center;
+  margin-left: 16px;
+  margin-right: 16px;
```
DmitriVanGuard: image
Misalignment. Here how it was before the change and in the design mock-up
image

src/components/Feed/Feed.js
haritonasty: isDesktop prop is for screen > than 768. For this case it isn't suitable, because on < 992px it should be in another place. Or I should create new var, that check width <= 992?

src/pages/Insights/InsightsPage.module.scss
haritonasty: because on mobile FeaturedInsights block has full screen width

src/components/Feed/Feed.js
DmitriVanGuard: Oh, I see. Sure, create a new var for now

src/components/Feed/Feed.module.scss
```diff
@@ -1,17 +1,118 @@
...
-}
+@import "../../mixins.scss";
 
 .date {
-    @include text-normal;
-    color: var(--grey-dark-2);
-    margin: 40px 0 16px;
+  @include text-normal;
+
+  color: var(--grey-dark-2);
```
Partysun: waterloo instead of grey-dark-2

Partysun:
Ok. Let's add storybook for this component.  @haritonasty 
I've fixed strorybook on our master. have to merge with newest ver..

src/components/FeaturedInsights/FeaturedInsightsHorizontal.js
```diff
@@ -0,0 +1,28 @@
...
+
+const FeaturedInsightsHorizontal = ({ maxLines, multilineTextId }) => (
+  <section className={styles.insights}>
+    <h4 className={styles.insights__title}>Featured insights</h4>
+    <div className={styles.insights__wrapper}>
+      <div className={styles.insights__scrollableWrapper}>
+        <div className={styles.insights__scrollable}>
+          <InsightsFeatured
+            maxLines={maxLines}
+            multilineTextId={multilineTextId}
```
Partysun: @haritonasty Do we really need this arg? What is it?

haritonasty: those props for multilineText component. In InsightsFeatured component it can be different (on mobile three lines, on desktop - two. I pass it form parent, but yes, I think I should set values here, not in parent.

haritonasty: @Partysun  no, I can't. I can do it only in FeaturedInsightsHorizontal, not in FIScrollable and FIWithTitle. They  can have different width on different pages, and multilineTextId depends on this.

PR #242, opened 2019-04-17
fixed liked state in small hearts

haritonasty:
image

ploblem desribed in [this task (third point)](https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/cffeb6e3eaf254588cb68a9a?card=San-4722)

waiting for Andrey response about design for likes, because it need refactor according with new design

### DIFF
--- a/src/components/Like/LikeBtn.module.scss
+++ b/src/components/Like/LikeBtn.module.scss
@@ -28,6 +28,7 @@
 
   & .icon {
     margin-right: 4px;
+    height: 12px;
   }
 }
 
--- a/src/components/Like/LikeBtn.module.scss
+++ b/src/components/Like/LikeBtn.module.scss
@@ -23,11 +23,12 @@
 }
 
 .small {
-  height: 12px;
   stroke-width: 0;
-  margin-top: -2px;
   fill: var(--casper);
-  margin-right: 4px;
+
+  & .icon {
+    margin-right: 4px;
+  }
 }
 
 .liked {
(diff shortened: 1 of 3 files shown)


src/components/Like/LikeBtn.module.scss
```diff
@@ -23,11 +23,13 @@
 }
 
 .small {
-  height: 12px;
   stroke-width: 0;
-  margin-top: -2px;
```
DmitriVanGuard: Without negative margin
image

src/components/Like/LikeBtn.module.scss
```diff
@@ -23,11 +23,13 @@
...
   fill: var(--casper);
-  margin-right: 4px;
+
+  & .icon {
```
DmitriVanGuard: CSS like this, without objective reason, should be avoided.
Amount of checks in js file didn't change meanwhile, here we have added a new conditional. `.small` should be applied directly to the `Icon`.

src/components/Like/LikeBtn.module.scss
```diff
@@ -23,11 +23,13 @@
 }
 
 .small {
-  height: 12px;
   stroke-width: 0;
-  margin-top: -2px;
   fill: var(--casper);
```
DmitriVanGuard: The only necessary change for fixing the bug was to move `fill: var(--casper);` inside the `grey` class

haritonasty: yes, it can be true, if we didn't have "like" without fill. So, I'm waiting Andrey for new desing for likes, he said, that likes should be another

PR #244, opened 2019-04-19
feature/sticky banner

haritonasty:
fixed styles for static banner and added new sticky banner
task [here](https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/cffeb6e3eaf254588cb68a9a?card=San-4683)

I decided to check only `top` property of static banner, because for check `bottom` - need know `window.innerHeight` and as I understand, reading this property [invoke trigger synchronous reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a)

1) upd banner on dashboard page with new design:
image

2) tablet version for static banner:
image

3) mobile version for static banner:
image

1) desktop version for sticky banner:
image

2) tablet version for sticky banner:
image

3) mobile version for sticky banner:
image


src/components/Banner/AnonBannerSticky.js
```diff
@@ -0,0 +1,58 @@
...
+  componentDidMount () {
+    this.viewportHeight = window.innerHeight
+  }
+
+  componentDidUpdate (prevProps) {
+    if (!prevProps.bannerStaticRef && this.props.bannerStaticRef) {
+      this.checkVisibility()
+      this.timer = setTimeout(this.checkVisibility, 50)
+    }
+  }
```
haritonasty: timer - small hack for new check after short period. I did it, because images in insights can loading, when banner has already mounted. This timer for them. I know, that when slow network it can be so short period, but onLoad() for image I can't use, because I didn't see in Editor this ability

haritonasty: @DmitriVanGuard @Partysun what do you think?

src/components/Banner/BannerMobile.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { Dialog } from '@santiment-network/ui'
+import styles from './BannerMobile.module.scss'
+
+const classes = { title: styles.title }
+
+const BannerMobile = ({ children }) => {
+  return (
+    <Dialog trigger={<></>} position='bottom' defaultOpen classes={classes}>
```
Partysun: @DmitriVanGuard Is it necessary to pass trigger for controlled behavior?

Partysun:
@haritonasty Try to make smaller PRs.) 200-300 it's enough usually.

src/pages/Insights/InsightViewPage.js
```diff
@@ -29,6 +30,14 @@ const InsightViewPage = ({
   isDesktop,
   isLoggedIn
 }) => {
+  const [bannerRef, setBannerRef] = useState()
+
+  const measuredRef = useCallback(node => {
```
DmitriVanGuard: `useRef` instead of `useCallback`

src/components/Banner/BannerMobile.js
DmitriVanGuard: This is a case for a `Modal` with a render callback as children. `Dialog` is an entity which has a straight connection to some action, e.g. click on the "delete" button, and following response to it.

src/components/Banner/BannerMobile.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { Dialog } from '@santiment-network/ui'
+import styles from './BannerMobile.module.scss'
+
+const classes = { title: styles.title }
```
DmitriVanGuard: `styles` is already an object which contains `title` property

src/components/Banner/AnonBannerSticky.js
```diff
@@ -0,0 +1,58 @@
...
+  componentWillUnmount () {
+    clearTimeout(this.timer)
+  }
+
+  checkVisibility = () => {
+    const { bannerStaticRef } = this.props
+
+    if (bannerStaticRef) {
+      const bannerTop = bannerStaticRef.getBoundingClientRect().top + 100
+      const isVisible = bannerTop <= this.viewportHeight && bannerTop >= 0
```
DmitriVanGuard: Static banner is always at the bottom of the page -> `bannerTop >= 0` is redundant

src/components/Banner/AnonBannerSticky.js
```diff
@@ -0,0 +1,58 @@
...
+
+  componentWillUnmount () {
+    clearTimeout(this.timer)
+  }
+
+  checkVisibility = () => {
+    const { bannerStaticRef } = this.props
+
+    if (bannerStaticRef) {
+      const bannerTop = bannerStaticRef.getBoundingClientRect().top + 100
```
DmitriVanGuard: What is this magic `100`?

src/components/Banner/AnonBannerSticky.js
```diff
@@ -0,0 +1,58 @@
...
+import { DesktopOnly, MobileOnly } from './../Responsive'
+import BannerDesktop from './BannerDesktop'
+import BannerMobile from './BannerMobile'
+import StickyBannerContent from './StickyBannerContent'
+
+class AnonBannerSticky extends React.PureComponent {
+  state = { isVisible: false }
+
+  componentDidMount () {
+    this.viewportHeight = window.innerHeight
```
DmitriVanGuard: What is the reason for creating this variable? Why it cannot be accessed directly via `window`?

src/components/Banner/AnonBannerSticky.js
DmitriVanGuard: There is no need for it.
As you pointed out, `50` milliseconds is too fast for slow connections. It's even fast for `4g`. In result, it gives us 0 benefits and 2 extra complexities: lifecycle method and timer.

src/components/Banner/AnonBannerSticky.js
```diff
@@ -0,0 +1,58 @@
...
+          <BannerDesktop
+            isVisible={this.state.isVisible}
+            checkVisibility={this.checkVisibility}
+          >
+            <StickyBannerContent />
+          </BannerDesktop>
+        </DesktopOnly>
+        {!this.state.isVisible && (
+          <MobileOnly>
+            <BannerMobile>
```
DmitriVanGuard: `<MobileOnly>` then `{!this.state.isVisible && ( ... `

src/components/Banner/AnonBannerSticky.js
haritonasty: ok, but if you open long insight with image and won't scroll - you won't see any banner, because data of top static banner will be incorrect. 150 - ok for 4g, I can change

src/pages/Insights/InsightViewPage.js
haritonasty: useRef not suitable, because it mutate ref, but I need to update data after getting ref. [Documentation](https://reactjs.org/docs/hooks-reference.html#useref)

src/components/Banner/BannerMobile.js
haritonasty: hm, ok. You wrote that `Modal` is  low-level component, so I decided use `Dialog`) I'll change

src/components/Banner/AnonBannerSticky.js
DmitriVanGuard: If it's a long insight user will scroll to see its content which will result in a banner popping up. This arbitrary value does not guaranty anything.

src/components/Banner/AnonBannerSticky.js
haritonasty: situation: when insight has all viewport height and static banner is showing only for 5-10px. I think user should see enough part of static banner. 
Or, if you mean that it should be as variable, I can't come up with a name for this number))

src/pages/Insights/InsightViewPage.js
DmitriVanGuard: And what's the problem with using the reference? `checkVisibility` still we able to get all necessary data for calculations

src/components/Banner/AnonBannerSticky.js
haritonasty: it use every time in callback for scroll. As I understand, reading this property [invoke synchronous reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a)

src/components/Banner/AnonBannerSticky.js
DmitriVanGuard: Sure, make sense 👍.
Proposed variable name: `BOTTOM_BANNER_EXTRA_VIEWPORT`

src/components/Banner/AnonBannerSticky.js
haritonasty: ok, I'll remove it

src/pages/Insights/InsightViewPage.js
haritonasty: because when checkVisibility is called - ref is empty. I tried useRef and after change to useCallback, because was problem with it

src/components/Banner/AnonBannerSticky.js
DmitriVanGuard: Nice catch 👍.
Speaking of performance, it's good to throttle the scroll event

haritonasty: I was thinking about it, I'll add it, ok)

src/pages/Insights/InsightViewPage.js
DmitriVanGuard: Don't know what was wrong for you, but it's working fine for me
image

haritonasty: it work because you didn't delete timer with `50` maybe

haritonasty: I'll try

DmitriVanGuard: Oh, so you wanted for it to appear from the beginning. It's okay if it will show up after the scroll

src/components/Banner/BannerMobile.js
haritonasty: I didn't understand( 
You mean write: 
` <Dialog trigger={<></>} position='bottom' defaultOpen classes={styles.title}>`
 without creating object for that?

that doesn't work for me without object

src/pages/Insights/InsightViewPage.js
haritonasty: what about mobile? I don't use scroll event here

DmitriVanGuard: The problem is in the component structure. `BannerMobile` is somehow aware of the logic which is applied to the `BannerDesktop`. That's an incorrect usage of the composition.
`AnonBannerSticky` should render `BannerMobile` if it's a mobile device and `AnonBannerStickyDesktop` if it's a desktop. `AnonBannerStickyDesktop`would be a component, which contains necessary logic for the conditional rendering of the `BannerMobile` desktop.

haritonasty: hm.. you mean that on mobile I should show sticky banner always when insight opened? Without checking position of static banner and it's OK if they will be both on screen?

DmitriVanGuard: @Partysun Didn't we decided to show banner by default after insight was opened on a mobile device?

src/components/Banner/BannerMobile.js
DmitriVanGuard: `classes={styles}`

haritonasty: oh, yes, my mistake

src/components/Banner/AnonBanner/AnonBannerB.module.scss
```diff
@@ -1,29 +1,32 @@
-@import '~@santiment-network/ui/variables.scss';
+@import '../../../../node_modules/@santiment-network/ui/variables';
```
DmitriVanGuard: Absolute import should be used for the modules

src/components/Banner/AnonBanner/AnonBanner.module.scss
```diff
@@ -0,0 +1,103 @@
+@import '../../../../node_modules/@santiment-network/ui/variables';
+@import '../../../mixins';
+
+.banner {
```
DmitriVanGuard: Current styles do not match base styles in [master](https://github.com/santiment/app/blob/a5ea90525cc3f631177f5641646c699d80ebde0b/src/components/AnonBanner/AnonBanner.module.scss#L1)

src/components/Banner/AnonBanner/AnonBannerA.module.scss
```diff
@@ -0,0 +1,7 @@
+@import '../../../../node_modules/@santiment-network/ui/variables';
+
```
DmitriVanGuard: Current styles do not match base styles in [master](https://github.com/santiment/app/blob/a5ea90525cc3f631177f5641646c699d80ebde0b/src/components/AnonBanner/AnonBannerA.module.scss#L2)

src/components/Banner/AnonBanner/AnonBanner.module.scss
haritonasty: yes, because Yura requested make static banner adaptive for tablet and mobile

DmitriVanGuard: @haritonasty I'm not speaking about responsive mixins. `__top`, `__description` and so on differ from the master branch, which has a clear separation between different variants

haritonasty: ok, done)

PR #245, opened 2019-04-19
added sortBy hypeScore

haritonasty:
image

### DIFF
--- a/src/components/Trends/GetHypedTrends.js
+++ b/src/components/Trends/GetHypedTrends.js
@@ -4,10 +4,6 @@ import { compose, pure } from 'recompose'
 import * as actions from './actions.js'
 import { simpleSort } from '../../utils/sortMethods'
 
-export const SORT_TYPES = {
-  hypeScore: 'score'
-}
-
 class GetHypedTrends extends React.Component {
   render () {
     const { render, items, ...rest } = this.props
@@ -25,7 +21,7 @@ class GetHypedTrends extends React.Component {
 }
 
 const sortByHype = items => {
-  const sortedData = items[0].topWords.sort(sort(SORT_TYPES.hypeScore))
+  const sortedData = items[0].topWords.sort(sort('score'))
   return {
     ...items,
     topWords: sortedData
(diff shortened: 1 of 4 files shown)


src/components/Trends/GetHypedTrends.js
```diff
@@ -2,18 +2,34 @@ import React from 'react'
...
 
+const sortByHype = items => {
+  const sortedData = items[0].topWords.sort(sort('score'))
+  return {
+    ...items,
+    topWords: sortedData
+  }
+}
+
+const sort = sortBy => (a, b) => simpleSort(a[sortBy], b[sortBy])
```
Partysun: Move this to util file. @haritonasty

Partysun: And find one more place in our code, where we have this function and replace it with util method.

src/components/Trends/GetHypedTrends.js
```diff
@@ -2,18 +2,34 @@ import React from 'react'
...
 
 class GetHypedTrends extends React.Component {
   render () {
-    const { render, ...rest } = this.props
-    return render(rest)
+    const { render, items, ...rest } = this.props
+    const trends = items.slice(-1)
+    const props = {
+      ...rest,
+      items: trends.length > 0 ? [sortByHype(trends)] : trends
```
Partysun: I don't understand this case. If trends.length <= 0, trends === []
why did you write ? ... : trends ?

Also you return only one array from items. What were items before we returned?
Can we create text for this.

haritonasty: lat question: https://github.com/santiment/app/blob/feature/hype-score-sort/src/components/Trends/TrendsTable/TrendsTables.js

this component expect array ("trends" prop), but it really not nesessary, as I understand. Should I rewrite this logic here?
inside array: datetime, topWords, __typename fileds

src/components/Trends/GetHypedTrends.js
```diff
@@ -2,18 +2,32 @@ import React from 'react'
...
```
DmitriVanGuard: This line breaks the logic of multiple trend snapshots. If `trends.length > 0 ?` is true, then `items` will always be a one item array

haritonasty: oh, yes, now I understand for what was 3 arrays

PR #249, opened 2019-04-21
fix/console-error

haritonasty:
PR fixed this error
image

after upd `react-sizes` also fixed this one
photo_2019-04-24_11-03-48


Partysun:
@DmitriVanGuard Check Smooth. If it's ok we can merge this.

PR #252, opened 2019-04-22
enhance/more-project-details

haritonasty:
added change7d parameter:
image


src/components/ValueChange/ValueChange.js
```diff
@@ -10,13 +10,14 @@ const Change = {
 
 const notChanged = ['texas-rose', 'lock-small']
 
-const ValueChange = ({ className, change, render }) => {
+const ValueChange = ({ className, change, render, label }) => {
```
DmitriVanGuard: **[1]** `ValueChange` is render-agnostic and should be only aware of **changed value**. If some kind of element should be rendered close it, then just make it as a sibling

src/pages/Detailed/DetailedHeader.js
```diff
@@ -57,10 +57,18 @@ const DetailedHeader = ({
...
+        <div>
+          <PercentChanges
+            className={styles.percentChanges}
+            changes={project.percentChange24h}
+            label='24h'
+          />
+          <PercentChanges
+            className={styles.percentChanges}
+            changes={project.percentChange7d}
+            label='7d'
```
DmitriVanGuard: See **[1]**. Render label as `<Label accent='...'>...</Label>`

src/pages/Detailed/DetailedHeader.module.css
```diff
@@ -68,11 +68,15 @@
 }
 
 .percentChanges {
-  font-size: 16px;
+  font-size: 14px;
   font-weight: 300;
   text-align: right;
 }
 
+.percentChanges:first-of-type {
```
DmitriVanGuard: Use scss syntax. `&:first-of-type`

src/pages/Detailed/DetailedHeader.js
```diff
@@ -9,7 +9,8 @@ import PercentChanges from './../../components/PercentChanges'
 import WatchlistsPopup from './../../components/WatchlistPopup/WatchlistsPopup'
 import ChooseWatchlists from './../../components/WatchlistPopup/ChooseWatchlists'
 import { formatNumber } from './../../utils/formatting'
-import styles from './DetailedHeader.module.css'
+import { Label } from '@santiment-network/ui'
```
DmitriVanGuard: "Module imports" should be placed before "path imports"

PR #255, opened 2019-04-25
feature/ab-tests-setup

DmitriVanGuard:
#### Summary
PR includes:
- Initial setup for the `A/B` tests;
- Experiments are connected to the **Mixpanel**;
- `A/B` test on the `AnonBanner` component.
- Proposed code structure when dealing with `A/B` tests.

#### Code structure
When some component should be tested, multiple files should be created for each experiment variant and lazy loaded via `react-loadable`. Variant separation is also true for styles.
I decided to propose this approach mainly because it's really hard to have one dynamic component for different variants. Often, variants will be significantly different in markup and applied styles. By taking this approach, after the end of the experiment, we can delete every file but the chosen one, which is much easier than transforming dynamic component into static one and dealing with its styles. Another benefit is an easiness in which we can modify each variant separately.

#### Screenshots
- AnonBanner-A
image

- AnonBanner-B
image


src/components/AnonBanner/AnonBannerExperiment.js
```diff
@@ -0,0 +1,59 @@
...
+          <AnonBannerB {...this.props} onClick={this.onClick} />
+        </Variant>
+      </Experiment>
+    )
+  }
+}
+
+// Called when the experiment is displayed to the user.
+emitter.addPlayListener(function (Experiment, Variant) {
+  console.log(`Displaying experiment "${Experiment}" variant "${Variant}"`)
```
Partysun: Clean

PR #261, opened 2019-04-29
feature/news

haritonasty:
**Created News component for TrendsExplorePage**

image

image

PR #267, opened 2019-05-10
feature/project-news

haritonasty:
**It should be merged after [this](https://github.com/santiment/app/pull/261) and [this](https://github.com/santiment/app/pull/266)**

 - Created small Card for news
- Added them to project card on mobile

image


src/components/News/NewsSmallCard.js
```diff
@@ -0,0 +1,37 @@
...
+const NewsSmallCard = ({ title, description, sourceName, url, datetime }) => {
+  return (
+    <a
+      href={url}
+      target='_blank'
+      rel='noopener noreferrer'
+      className={cx(styles.link)}
+    >
+      <div className={styles.news}>
+        <div className={styles.news__title}>
```
DmitriVanGuard: `styles` object incapsulates CSS only for the current component. So we can safely avoid BEM to reduce the number of necessary characters.
-> `.title`, `.date` and so on

src/components/News/NewsSmallCard.js
```diff
@@ -0,0 +1,37 @@
...
+
+const NewsSmallCard = ({ title, description, sourceName, url, datetime }) => {
+  return (
+    <a
+      href={url}
+      target='_blank'
+      rel='noopener noreferrer'
+      className={cx(styles.link)}
+    >
+      <div className={styles.news}>
```
DmitriVanGuard: `a` could be a single top-wrapper (without the `div` on LOC `15`), which will help us to avoid extra node creations

PR #268, opened 2019-05-13
enhance/top50erc20-new-api

haritonasty:
**Rebuilt top 50 erc20 projects with new API**

[favro card](https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/cffeb6e3eaf254588cb68a9a?card=San-4891)

### DIFF
--- a/src/components/WatchlistPopup/WatchlistGQL.js
+++ b/src/components/WatchlistPopup/WatchlistGQL.js
@@ -44,6 +44,15 @@ export const publicWatchlistGQL = gql`
   }
 `
 
+export const projectsByFunctionGQL = gql`
+  query allProjectsByFunction($function: json!) {
+    allProjectsByFunction(function: $function) {
+      id
+      slug
+    }
+  }
+`
+
 export const FEATURED_WATCHLIST_QUERY = gql`
   query featuredWatchlists {
     featuredWatchlists {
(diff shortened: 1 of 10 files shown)


src/epics/fetchAssetsEpic.js
```diff
@@ -240,3 +241,63 @@ export const fetchAssetsFromSharedListEpic = (action$, store, { client }) =>
...
+            payload: {
+              items: [],
+              isLoading: false,
+              error: false
+            }
+          })
+        }
+
+        return Observable.forkJoin(queries)
+          .delayWhen(() => Observable.timer(500 + startTime - Date.now()))
```
DmitriVanGuard: Is an arbitrary delay so necessary? Because it will cause a blink even when the data was cached.

src/pages/assets/GetAssets.js
```diff
@@ -24,33 +25,40 @@ class GetAssets extends Component {
...
+  getInfoFromListname = (listname = '') => {
+    const [listName, listId] = listname.split('@')
+    const listByFunction = WATCHLISTS_BY_FUNCTION.find(
+      ({ assetType }) => assetType === listName
+    )
+    return listId
+      ? { listName, listId }
+      : {
+        listName,
+        listId,
```
DmitriVanGuard: If the `listId` is falsy we still add it to the object. Why?

src/components/WatchlistPopup/WatchlistGQL.js
```diff
@@ -44,6 +44,15 @@ export const publicWatchlistGQL = gql`
   }
 `
 
+export const projectsByFunctionGQL = gql`
```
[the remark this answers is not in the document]
haritonasty: but previous gql templates (in this file) with this style. Maybe rename the last? Or we decided to follow `_QUERY` style guide?

PR #269, opened 2019-05-13
fix/erc20

haritonasty:
**1. Added info in `assets/` to `erc20` card**
image

**2. Added comparing on graphic**
image

### DIFF
--- a/src/components/TotalMarketcapWidget/GetTotalMarketcap.js
+++ b/src/components/TotalMarketcapWidget/GetTotalMarketcap.js
@@ -26,6 +26,10 @@ const getMarketcapQuery = (type, projects) => {
     })
   })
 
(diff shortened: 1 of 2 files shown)


src/components/TotalMarketcapWidget/GetTotalMarketcap.js
```diff
@@ -26,10 +26,33 @@ const getMarketcapQuery = (type, projects) => {
     })
   })
 
-  if (type !== 'list') {
+  const slugsQueryERC20 = graphql(totalMarketcapGQL, {
```
DmitriVanGuard: In the case when `type` equals to `all` `slugsQueryERC20`variable will be initialized but not used. Better to move it after first return statement (LOC 47)

PR #271, opened 2019-05-16
fix/link-retry

haritonasty:
fixed retry logic with [documentation example](https://www.apollographql.com/docs/link/links/retry)

### DIFF
--- a/src/apollo/retry-link.js
+++ b/src/apollo/retry-link.js
@@ -1,14 +1,13 @@
 import { RetryLink } from 'apollo-link-retry'
 
 const retryLink = new RetryLink({
-  delay: {
-    initial: 500,
-    max: Infinity,
-    jitter: true
-  },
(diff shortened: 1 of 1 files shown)


src/apollo/retry-link.js
```diff
@@ -1,14 +1,13 @@
...
-  delay: {
-    initial: 500,
-    max: Infinity,
-    jitter: true
-  },
   attempts: (count, operation, error) => {
-    const max = operation.getContext().maxAttempts || 10000
-    return !!error && count < max && operation.getContext().isRetriable
+    const { isRetriable, maxAttempts } = operation.getContext()
+    const max = maxAttempts || 100
```
DmitriVanGuard: `100` feels like a potential DDoS. Maybe to lower it to `10`?

haritonasty: agree

PR #280, opened 2019-05-28
fix/volume-to-marketcap

haritonasty:
**Page: category/watchlist page ([example](http://localhost:3000/assets/erc20)) **

[favro card](https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/cffeb6e3eaf254588cb68a9a?card=San-5014)

**Changed volume for all assets to list marketcap**

**Before:**
image
**Now:**
image

### DIFF
--- a/src/components/TotalMarketcapWidget/TotalMarketcapWidget.js
+++ b/src/components/TotalMarketcapWidget/TotalMarketcapWidget.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React from 'react'
 import cx from 'classnames'
 import {
   Area,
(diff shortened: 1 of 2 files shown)


src/components/TotalMarketcapWidget/TotalMarketcapWidget.js
```diff
@@ -38,48 +32,34 @@ const tickFormatter = date => {
 }
 
 class TotalMarketcapWidget extends Component {
-  state = {
```
DmitriVanGuard: If the component is free of `state` then we could refactor it as a function

PR #288, opened 2019-05-31
fix/facebook-share

haritonasty:
[Favro card](https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/cffeb6e3eaf254588cb68a9a?card=San-5006)

Problem happened because `metrics` array was missed and metrics returned as separate arrays (on screenshot `metrics` include `price` by default):
image

Result:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "ms": "^2.1.1",
     "nprogress": "^0.2.0",
     "outliers": "^0.0.3",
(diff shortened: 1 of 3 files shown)


DmitriVanGuard:
@haritonasty One possible solution is to make a small change to the url parser. We can change `array` format to one of the followings and facebook will not escape array brackets. 
image

PR #290, opened 2019-05-31
fix/search

haritonasty:
[favro card](https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/cffeb6e3eaf254588cb68a9a?card=San-4954)

my suggest, that search didn't work for `TRON`, because we used only `includes()` method and the first 5 projects includes this substring. TRON project also is searching, but didn't include in first five ones

we can't test it on stage, because set of projects isn't the same as on prod

### DIFF
--- a/src/components/Search/SearchProjects.js
+++ b/src/components/Search/SearchProjects.js
@@ -15,6 +15,8 @@ const SearchProjects = ({ data: { allProjects = [] }, ...props }) => {
       predicate={searchTerm => {
         const upperCaseSearchTerm = searchTerm.toUpperCase()
         return ({ ticker, name }) =>
+          name.toUpperCase() === upperCaseSearchTerm ||
+          ticker.toUpperCase() === upperCaseSearchTerm ||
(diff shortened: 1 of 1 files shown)


DmitriVanGuard:
@haritonasty Proposal TL;DT: suggestions should be sorted by the length of the name extra characters in the ascending order.
Here is the idea in details. 
First of all, when a user searches the asset by ticker (like `TRX`), it's guaranteed that the result will contain the correct one. Meanwhile, things are complicated when we are dealing with names (that's why we have this pr in the first place). In the case of `TRON` there also are `bioTRON`, `paTRON` and so on. Of course, the first suggestion should be the asset with a most matched character percentage (e.g `TRON` has `4/4` matched characters, `bioTRON` - `4/7`). In the sort function, the search string in the asset name could be replaced with the `''` and then the comparison between the length of modified strings is made.

```js
const sortByMatchLength = ({name: aName}, {name: bName}) => {
  return aName.toUpperCase().replace(searchTerm, '') - bName.toUpperCase().replace(searchTerm, '');
}
```

PR #330, opened 2019-06-17
Feature/san 5100 wallets trigger

GarageInc:
Return chart in trigger form


Partysun:
@GarageInc Is it ready PR? Can you add screenshots and list of changes, pls.

PR #352, opened 2019-06-21
5100: fixed charts for DAA in triggers

GarageInc:
1.Fixed bug with DDA charts (not showed)
image
2. Add delete dialog
image


src/ducks/SANCharts/utils.js
```diff
@@ -34,7 +34,10 @@ export const Metrics = {
   dailyActiveAddresses: {
     node: Bar,
     color: 'malibu',
-    label: 'Daily Active Addresses'
+    label: 'Daily Active Addresses',
+    dataKey: 'active_addresses',
+    orientation: 'right',
```
Partysun: Is it critical changes for our /charts ?

GarageInc: Crytical. Will change to custom active_addresses back

GarageInc: Changed

PR #376, opened 2019-07-02
Fix/san 5100

GarageInc:
Trigger form: removed borders, add price change component


src/ducks/Signals/signalFormManager/signalCrudForm/formParts/TriggerLastPrice.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { formatNumber } from '../../../../../utils/formatting'
+import styles from './../signal/TriggerForm.module.scss'
+
+export const LastPriceComponent = ({ lastPrice }) => {
+  if (lastPrice === undefined) return ''
+
+  return (
+    <div className={styles.lastPrice}>
+      <span>
```
Partysun: Why span inside div?

PR #380, opened 2019-07-02
Feature/san 5100 v2

GarageInc:
1. Preview chart
image

2. Small popup with "...More"
image

3. Signal settings step route(fastest way)
image


src/components/SignalCard/SignalCard.module.scss
```diff
@@ -52,16 +52,76 @@
...
+}
+
+.popup {
+  position: absolute;
+  left: 0;
+  top: 25px;
+  background: var(--white);
+
+  /* N200 - Porcelain */
+  border: 1px solid var(--porcelain);
```
haritonasty: LOC 73 - you can delete it

src/components/SignalCard/SignalCard.module.scss
```diff
@@ -52,16 +52,76 @@
...
+
 .bottom {
   margin-top: 19px;
   display: flex;
-  justify-content: space-between;
 
   &__right {
     display: flex;
   }
 }
```
haritonasty: LOC 91 - 98 are still used or no?

src/components/SignalCard/SignalCard.module.scss
GarageInc: done

src/components/SignalCard/SignalCard.module.scss
GarageInc: Will clear

PR #383, opened 2019-07-02
feature/mobile-search

haritonasty:
**Summary**
- added ability to search trends on mobile
- added assets and trends tabs to search page on mobile

**WARNING** Pls, merge it after [that](https://github.com/santiment/app/pull/386)

**Screenshots**
image
image


src/pages/SearchMobilePage/SearchMobilePage.js
```diff
@@ -1,21 +1,108 @@
...
+  const [selectedTab, selectTab] = useState(TABS[0].index)
+  const onSelectTab = selected => selectTab(selected)
+  const assetsRecent = getRecentAssets()
+  const assetsWithInfo = assetsRecent.map(asset => ({
+    ...allProjects.find(({ slug }) => slug === asset)
+  }))
+  const [assets, setAssets] = useState(assetsWithInfo)
+  const [trends, setTrends] = useState(getRecentTrends())
+
+  console.log(trends, assets, selectedTab)
```
Partysun: Clean logs

src/pages/SearchMobilePage/SearchMobilePage.js
```diff
@@ -1,21 +1,108 @@
...
+    content: 'Trends'
+  }
+]
+
+const SearchMobilePage = ({ history }) => {
+  const [selectedTab, selectTab] = useState(TABS[0].index)
+  const onSelectTab = selected => selectTab(selected)
+  const assetsRecent = getRecentAssets()
+  const assetsWithInfo = assetsRecent.map(asset => ({
+    ...allProjects.find(({ slug }) => slug === asset)
```
Partysun: If the user opens the project which not contains in saved json, then we can't show it in recent. It's ok now, but we need to save this as a card

src/pages/SearchMobilePage/SearchMobilePage.js
```diff
@@ -1,21 +1,108 @@
...
+  }))
+  const [assets, setAssets] = useState(assetsWithInfo)
+  const [trends, setTrends] = useState(getRecentTrends())
```
Partysun: Why did not use assetsRecent const, which you define above? @haritonasty

src/pages/SearchMobilePage/SearchMobilePage.js
```diff
@@ -1,21 +1,108 @@
...
+      <Tabs
+        options={TABS}
+        defaultSelectedIndex={selectedTab}
+        onSelect={onSelectTab}
+        className={styles.tabs}
+      />
+      <div className={styles.recentWrapper}>
+        <h3 className={styles.caption}>Recently searched</h3>
+        {selectedTab === TABS[0].index &&
+          assets.map(({ name, ticker, coinmarketcapId }) => (
```
Partysun: Can we use our slug instead cmcID?

src/pages/SearchMobilePage/SearchMobilePage.js
Partysun: Do we have a limit of size for recent Trends/assets? @haritonasty

haritonasty: yes, 5

src/pages/SearchMobilePage/SearchMobilePage.js
haritonasty: I think, better to remove this logic now and use only slugs in recent block

PR #389, opened 2019-07-03
fix/filter-empty-recent

haritonasty:


src/pages/SearchMobilePage/SearchMobilePage.js
```diff
@@ -27,8 +27,8 @@ export const TABS = [
 const SearchMobilePage = ({ history }) => {
   const [selectedTab, selectTab] = useState(TABS[0].index)
   const onSelectTab = selected => selectTab(selected)
-  const [assets, setAssets] = useState(getRecentAssets())
-  const [trends, setTrends] = useState(getRecentTrends())
+  const [assets, setAssets] = useState(getRecentAssets().filter(Boolean))
+  const [trends, setTrends] = useState(getRecentTrends().filter(Boolean))
```
DmitriVanGuard: Hmm, strange. `saveRecents` should not save and `non-true` value [because of the native filtering](https://github.com/santiment/app/blob/c96f1b6014f4c833d280030dfba9f977bbd9a837/src/utils/recent.js#L17)

PR #397, opened 2019-07-03
5100: fixed trigger modal step "About"

GarageInc:
Done:
* step "About" in mobile view for trigger
image


haritonasty:
can you pls change in future mobile modal view as in another places, ex:
image

PR #402, opened 2019-07-05
fix/overflow-tags

haritonasty:
**Summary**
1) don't select more than 5 tags on trends tables
2) don't let to type 6th tag in insight
3) moved tags to new prePublish layout
4) fixed missed default (trending) tags (it was ok early, because selector mounted, but now it's in ContextMenu)
5) added notification after publishing

https://gifyu.com/image/ElS7

**Screenshots**
image


src/components/Trends/TrendsTable/TrendsTable.js
```diff
@@ -215,10 +216,11 @@ class TrendsTable extends PureComponent {
     const tableData = trendWords.map(({ word, score }, index) => {
       const [oldVolume = 0, newVolume = 0] = volumeChange[word] || []
       const isWordSelected = selectedTrends.has(word)
+      const isLimitReached = selectedTrends.size > 4 && !isWordSelected
```
DmitriVanGuard: Without a context it's really hard to identify "what limit is reached" just by reading the variable name. Proposed name: `hasMaxWordsSelected`

src/components/Insight/InsightEditor/InsightEditor.js
```diff
@@ -31,6 +31,10 @@ class InsightEditor extends Component {
     isEditing: false
   }
 
+  componentDidUpdate ({ tags }, { tags: stateTags }) {
+    if (tags.length > 0 && this.props.tags !== tags) this.setState({ tags })
+  }
```
DmitriVanGuard: `stateTags` is not being used.
@haritonasty please, can you explain why this lifecycle method is needed?

src/components/Insight/InsightEditor/InsightEditor.js
```diff
@@ -105,8 +109,9 @@ class InsightEditor extends Component {
   )
 
   render () {
-    const { id, title, tags, updatedAt, isUpdating, publishDraft } = this.props
+    const { id, title, updatedAt, isUpdating, publishDraft } = this.props
     const { isEditing } = this.state
+    const tags = [...new Set([...this.props.tags, ...this.state.tags])]
```
DmitriVanGuard: `this.props.tags` will change during the component lifetime?

src/components/Insight/InsightEditor/InsightEditorBottom.js
```diff
@@ -16,39 +18,56 @@ const InsightEditorBottom = ({
   hasMetTextRequirements
 }) => {
   const options = { from: new Date(updatedAt) }
+  const hasRequirements =
```
DmitriVanGuard: I guess, semantically better to call it `hasMatchedRequirements`

src/components/Insight/InsightEditor/InsightEditorBottom.js
```diff
@@ -16,39 +18,56 @@ const InsightEditorBottom = ({
...
+    hasMetTextRequirements.title && hasMetTextRequirements.text
+  const tags = defaultTags.filter(
+    ({ name }) => !name.endsWith('-trending-words')
```
DmitriVanGuard: [Existing `filter` function](https://github.com/santiment/app/blob/6f5bebce675cea48e9d50b4ffb6e47b67c3ced31/src/components/Insight/utils.js#L1)

src/components/Insight/InsightEditor/InsightEditor.js
haritonasty: image
sometimes new.props and new.state are empty and only from prev props we can restore tags(

src/components/Insight/InsightEditor/InsightEditor.js
haritonasty: yep. very often. I guess, it's because we use `[trends] SELECTED_WORDS_CLEAR` when draft insight in editing. It's (as I see) for our trends page, where we chose words

PR #406, opened 2019-07-06
feature/chart-interval-selector

DmitriVanGuard:
### Summary
Adding interval selector for a chart tool.
Available intervals are calculated according to the difference between `to` and `from` dates in days. Calculation is taken place [here](https://github.com/santiment/app/compare/enhance/chart-tool?expand=1#diff-a3ad42538f36ea4f3aac5d5abc1e7df5R19).

**We should decide what are appropriate predefined intervals for specific timeranges.**
- PR #408 should be merged first.

### Screenshots
video


haritonasty:
for me - `1h`, `12h`, `1d`, `2d`. They are so detailed and looks nice

DmitriVanGuard:
My suggestion is to keep `70-200` data points per time range

Partysun:
We need to auto setup smaller interval for user's time frame. To avoid not intresting chart)

PR #410, opened 2019-07-06
enhance/like new styles

haritonasty:
**Summary**
Added new styles for like 

**Screenshots**
in action

### DIFF
--- a/src/components/Insight/InsightCardInternals.js
+++ b/src/components/Insight/InsightCardInternals.js
@@ -30,7 +30,7 @@ const InsightCardInternals = ({
   votedAt,
   onLike,
   withAuthorPic,
-  infoOnly,
+  disabled,
   className,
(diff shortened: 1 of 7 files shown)


DmitriVanGuard:
My question is: Do we really need a `pulse` animation + JS listener and several conditions?
Here is the same thing written in plain CSS which follow the design library specification.
Only css like

image

haritonasty:
@DmitriVanGuard for question with pulse animation: on desktop not all users use clicks - I use tab. And I can't see animation if we will use :active

DmitriVanGuard:
Until this moment, our likes were completely "unfriendly" for a `tab` users. It's a `div` without a `tabindex` attribute. Anyway, support for such case is easy to implement: we need to just add `:focus` next to `:active`

PR #416, opened 2019-07-09
5100: add eth address validation for trigger form

GarageInc:
Done:
* ETH address checking on trigger form
image

### DIFF
--- a/src/ducks/Signals/utils/utils.js
+++ b/src/ducks/Signals/utils/utils.js
@@ -34,7 +34,7 @@ import {
   TIME_WINDOW_UNITS,
   getDefaultTimeRangeValue
 } from './constants'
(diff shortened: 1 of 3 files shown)


src/ducks/Signals/utils/utils.js
```diff
@@ -34,7 +34,7 @@ import {
   TIME_WINDOW_UNITS,
   getDefaultTimeRangeValue
 } from './constants'
-import { capitalizeStr } from '../../../utils/utils'
+import { capitalizeStr, isEthStrictAddress } from '../../../utils/utils'
```
IvanIvanoff: Is this doing the checksum (upper and lower case matter), too? Will it accept all lowercase or all uppercase because the backend accepts such addresses

PR #434, opened 2019-07-13
fix/chart-payment-wall

DmitriVanGuard:
### Summary
Showing `Paywall` on the chart;

### Screenshots
image


DmitriVanGuard:
@Partysun should `Paywall` affect `priceUsd` metric data visibility? Currently, all datapoints are visible

Partysun:
@DmitriVanGuard Not, let's show data, which available. But we can show info about paywall in the tooltip. Ex. You can' t see a lot of our data for more than 3 months until you upgrade a plan. For better text, we need to ask Ibis.

PR #446, opened 2019-07-17
fix/username-notification

haritonasty:
**Summary**
- Moved warning about missed username to notification (*once a week*)
- Removed translateY for `NotificationStack`, because we have notification with different heights and 
`--y-offset`: `calc(-${i}00% - ${i}0px)` we can't use. In additional, I've changed order for notifications, (because I deleted smooth moving down) and with inverse order and more that 1 notification not so bad styles
- upd notification styles for mobile

**Screenshots**
was:
image
now:
image


Partysun:
@haritonasty Can we show this only on insights page. Cause name we need only for insights.

src/pages/Insights/index.js
```diff
@@ -28,7 +31,29 @@ const LoadableUnAuthPage = Loadable({
...
+  showUsernameAlert,
+  hasUsername
+}) => {
+  const setLastShowUsernameAlert = () => {
+    showUsernameAlert()
+    localStorage.setItem('withoutUsernameSince', new Date())
+  }
+
+  if (isLoggedIn && !hasUsername) {
+    console.log('here')
```
DmitriVanGuard: Logging

PR #457, opened 2019-07-19
enhance/user-settings

haritonasty:
**Summary**
Added saving for user night/beta mode:
1) Use settings from `LS` firstly
2) If user doesn't have any keys for such settings in `LS` => pick up them from server
3) As we've already have fileds for such settings in `Local Storage` and and suddenly not reset user preferences => I create additional epics for this case: 
If `isNightModeEnabled` and `isBetaModeEnabled` = `true` in `LS` -> send them to server (if user logged in)


src/actions/types.js
```diff
@@ -7,7 +7,9 @@ export const APP_SHOW_ONBOARDING = '[app] SHOW_ONBOARDING'
 export const APP_GDPR_FAILED = '[app] GDPR_FAILED'
 export const APP_USER_HAS_INACTIVE_TOKEN = '[app] USER_HAS_INACTIVE_TOKEN'
 export const APP_USER_NIGHT_MODE_SAVE = '[app] USER_NIGHT_MODE_SAVE'
+export const APP_USER_NIGHT_MODE_SAVE_ERROR = '[app] USER_NIGHT_MODE_ERROR'
 export const APP_USER_BETA_MODE_SAVE = '[app] USER_BETA_MODE_SAVE'
+export const APP_USER_BETA_MODE_SAVE_ERROR = '[app] USER_BETA_MODE_SAVE_ERROR'
```
DmitriVanGuard: To preserve consistency across the codebase, it's better to end `error` name with a [`FAILED` postfix](https://github.com/santiment/app/pull/457/files#diff-fbed0602d73913936d71c47e897c48c5R69)

src/epics/handleBetaModeToggle.js
```diff
@@ -1,38 +1,92 @@
...
+          .mergeMap(({ data: { updateUserSettings: { isBetaMode } } }) => {
+            return value
+              ? Observable.of({ type: APP_USER_BETA_MODE_SAVE, payload: true })
+              : Observable.from([
+                { type: APP_USER_BETA_MODE_SAVE, payload: false },
+                { type: APP_USER_NEWS_SAVE, payload: false }
+              ])
+          })
+          .catch(handleErrorAndTriggerAction(APP_USER_BETA_MODE_SAVE_ERROR))
+      } else {
```
DmitriVanGuard: `else` is not necessary, because of the [`return` statement inside `if` block](https://github.com/santiment/app/pull/457/files#diff-0e20b76b28777795df84b9091165cf10R37) (`LOC 37`)

src/reducers/root-ui.js
```diff
@@ -1,10 +1,22 @@
...
+const isNightMode = loadKeyState('isNightMode')
+const isNightModeDeprecated = loadKeyState('isNightModeEnabled')
+const isBetaMode = loadKeyState('isBetaMode')
+const isBetaModeDeprecated = loadKeyState('isBetaModeEnabled')
 const isNewsEnabled = loadKeyState('isNewsEnabled') || false
 
+let isNightModeEnabled = isNightMode || false
+if (isNightMode === undefined && isNightModeDeprecated !== undefined) {
+  isNightModeEnabled = isNightModeDeprecated
+}
```
DmitriVanGuard: We could simplify it to 1 line since `undefined` is a **falsy** value 
```js
const isNightModeEnabled = isNightMode || isNightModeDeprecated
```

src/epics/handleNightModeToggle.js
```diff
@@ -1,26 +1,100 @@
...
       })
+    })
+
+export const sendNightModeIfDiff = (action$, store, { client }) =>
+  action$
+    .ofType(CHANGE_USER_DATA)
+    .filter(
+      () =>
+        loadKeyState('isNightMode') === undefined ||
+        loadKeyState('isNightModeEnabled') === true
```
DmitriVanGuard: If it's possible, could we take the value of the `isNightMode` from our "single source of truth" - redux store?
Redux store is initialised with a [`LS` value](https://github.com/santiment/app/pull/457/files?file-filters%5B%5D=.js#diff-e6c10270750f429a2a3409d7f330a028R4), so the setting value should be correct after the app has launched.

src/reducers/root-ui.js
haritonasty: The problem here, that we don't want to use `isNightModeDeprecated`, if `isNightMode = false`

src/epics/handleNightModeToggle.js
haritonasty: Here the case, when previous `LS` field (`isNightModeEnabled`) contain true statement and new field haven't determined yet (`isNightMode`) and we need to upd settings on server for user (for preventing diff between predefined `night mode` in LS and `default` on server after merge this PR). It's for consistency. 
In store we have only one field for night mode, that can contain value from any these two fields.
I can create two additional fields in store, but...I think it will be more confusion...I can write a NOTE here for what I use this epic.
___
After your review I also found a mistake, that here should be `&&` instead `||`. Thank you)

PR #462, opened 2019-07-23
enhance/watchlist-settings

haritonasty:
**Summary**
- **added ability to save hidden columns:**
- Initialized `redux-store` with default settings
- If we've already have settings saved in `LS` - also move them to `redux-store`. We need saving setting in `LS`, because: 1) not all watchlists have id (`all assets`, `erc20`, `top50erc20`), but we want to save settings for them too. 2) we can take settings from `LS` fast instead waiting it from server (in future maybe we wan't to ask server only required columns)
- When user logged in - fetch user watchlists settings after launch and save them to `redux-store`
- When user toggle columns - save to `LS` and to `redux-store`. If logged in - send to server also.
- Hid `ETH spent` but default for all watchlists and categories, except `erc20` and `top50erc20`. 
- `pageSize = 20` by default, except `top50erc20`, ehre `pageSize = 50`

*Note:* For now, I send to server all columns and hide them on client. If we'll add pagination `erc20` - all requests looks fast. 
But, If I'll send request with only required columns - when user will toggle them - user will be waiting. 
And also if we don't have setting for watchlist - in case where we ask only necessary columns we need to send request for settings before request for watchlist data.
We can combine maybe: ask only necessary and load in background remaining columns. If we don't have settings for watchlist - ask all and hide them on client.
What do you think, @Partysun, @DmitriVanGuard ?

**Draft** - API for watchlist settings is unstable currently.


Partysun:
@haritonasty Ask all and hide looks ok. for faster solution. But feels like a bad thing) We can use it, but I am talking with backend team about it.

src/pages/assets/AssetsTable.js
```diff
@@ -45,28 +49,60 @@ const AssetsTable = ({
...
+    changeHidden(savedHidden)
+    changeColumns(changeShowing(COLUMNS_SETTINGS, savedHidden))
+  }
+
+  const toggleColumn = ({ name, show, selectable }) => {
+    const toggledColumns = Object.assign({}, columns)
+    // NOTE(haritonasty): such access to the fields is necessary for Safari bug (shuffle properties)
+    toggledColumns[name] = {
+      ...toggledColumns[name],
       show: selectable ? !show : show
```
Partysun: @haritonasty why not show: !selectable

src/pages/assets/AssetsTable.js
```diff
@@ -45,28 +49,60 @@ const AssetsTable = ({
...
 
-    return changeColumnsSettings(toggledColumns)
+    if (selectable) {
+      const columns = show
+        ? [...savedHidden, name]
+        : savedHidden.filter(item => item !== name)
+      setHiddenColumns({
+        hiddenColumns: columns,
+        listId: typeInfo.listId,
+        listName
```
Partysun: move listName to the top. If we used this sort {c, d, a: asdf, b: sadf}

src/pages/assets/AssetsTable.js
haritonasty: we don't wan't toggle disabled items: 
image

PR #464, opened 2019-07-23
enhance/chart-paywall

DmitriVanGuard:
### Summary
- Updated design;
- Paywall is now aware of the zoom bounds;
- General speed increase (`4x` on small arrays, `>10x` on large arrays);

### Screenshots
image

### DIFF
--- a/src/ducks/SANCharts/Paywall.js
+++ b/src/ducks/SANCharts/Paywall.js
@@ -1,12 +1,7 @@
 import React from 'react'
 import { ReferenceArea } from 'recharts'
 import { binarySearch } from '../../pages/Trends/utils'
-import {
-  calculateUnitByFormat,
-  getTimeIntervalFromToday,
-  DAY,
-  MONTH
-} from '../../utils/dates'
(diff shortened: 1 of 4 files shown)


src/ducks/SANCharts/Paywall.js
```diff
@@ -0,0 +1,108 @@
...
+  return res && +new Date(res.datetime)
+}
+
+const displayPaywall = ({ data }) => {
+  if (!data || !data.length) return
+
+  const lastItemDate = new Date(data[data.length - 1].datetime)
+  const firstItemDate = new Date(data[0].datetime)
+
+  const { from: leftHistoricalDate } = getTimeIntervalFromToday(-3, MONTH)
```
Partysun: @DmitriVanGuard Better move interval as args.
For MVRV we will use very soon next logic. all history is free, last 3 or 1 month is blocked.

Partysun: For token age consumed and for MVrv we will have blocked only last 30 days.

PR #472, opened 2019-07-24
feature/chart-signal-creation

DmitriVanGuard:
### Summary
Allowing signal creation for a chosen slug.
### Screenshots
image
image
For anon users
image


Partysun:
Just "Create Signal"

PR #518, opened 2019-08-07
Update allProjects.json file with the latest data

IvanIvanoff:


src/allProjects.json
```diff
@@ -9267,6 +9813,20 @@
     "slug": "wings",
     "ticker": "WINGS"
   },
+  {
+    "coinmarketcapId": "wink-tronbet",
+    "id": "235738",
+    "name": "Wink",
+    "slug": "wink-tronbet",
+    "ticker": "WIN"
```
IvanIvanoff: I saw that this was missing the price in the insights: 
image

PR #525, opened 2019-08-09
Feature/check activities sw

GarageInc:
Done:
* service-worker which every N seconds/minutes check for new activities before user load activities page. If page loaded than code of page send message to service worker and it resets a counter of last checked activities.
Webpack build service worker from ./src-folder into ./public folder with required urls. Internal storage is IndexDb.

Title: ':N new activities in Sonar!'
Description: 'Open to check https://app-stage.santiment.net/sonar/activity'
image

Need to check all changes in codebase.

* add toggle in /account page for Web pushes
image
image

* fixed some styles in /account page(padding, margins, mobile styles)


Partysun:
We need to make a special screen when we will ask about permission access for notification.
Also console.log and production checking, you need to revisit it here.

Partysun:
> > We need to make a special screen when we will ask about permission access for notification.
> > Also console.log and production checking, you need to revisit it here.
> 
> @Partysun , do you mean such popup(as for cookies)?
> image

Nope. I asked Alan to do this screen.

PR #533, opened 2019-08-15
fix: new sonar service worker branch

GarageInc:
Done:
* service-worker which every N seconds/minutes check for new activities before user load activities page. If page loaded than code of page send message to service worker and it resets a counter of last checked activities.
Webpack build service worker from ./src-folder into ./public folder. Internal storage is IndexDb.
Title: ':N new activities in Sonar!'
Description: 'Open to check https://app-stage.santiment.net/sonar/activity'
image

* add toggle in /account page for Web pushes
image
image

* add support of 'Web Push' channel in trigger form
* fixed some styles in /account page(padding, margins, mobile styles)


GarageInc:
Created a new branch, old was deleted

PR #536, opened 2019-08-16
fix/min-volume

haritonasty:
**Summary**
- correct behavior for `minVolume` variable in all `all*` queries (pass it even if `=0`)
- now for `All assets` and for `ERC20` `minVolume = 10000` as was in code (but didn't work)
- changed `if !value` to `if parseFloat(value) ===  NaN` in assets table (because `0` showed as `no data`)  

**Screenshoots**
image

### DIFF
--- a/src/pages/assets/asset-columns.js
+++ b/src/pages/assets/asset-columns.js
@@ -15,6 +15,8 @@ const HeaderWithDesc = ({ description, heading }) => (
   </Tooltip>
 )
 
+const isValidValue = value => !isNaN(parseFloat(value))
+
 export const COLUMNS = preload => [
   {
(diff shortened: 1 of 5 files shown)


IvanIvanoff:
🎉 🎉 🎉

src/pages/assets/asset-columns.js
```diff
@@ -85,7 +89,11 @@ export const COLUMNS = preload => [
     }),
     Cell: ({ value: { change24h } }) => (
       <div className='overview-price-percent'>
-        {change24h ? <PercentChanges changes={change24h} /> : 'No data'}
+        {isValidValue(change24h) ? (
+          <PercentChanges changes={change24h} />
+        ) : (
+          'No data'
```
GarageInc: 'No data' in const variable?

PR #554, opened 2019-08-21
enhance/metrics-api

DmitriVanGuard:
### Summary
- Improvements of the `GetTimeSeries` component;
- Using `getMetric` endpoint for the `transaction_volume` metric. More metrics yet to come.

### What's changed?
- Required metrics are now passed as the array objects via [`metrics` prop](https://github.com/santiment/app/pull/554/files#diff-34bac4f2ff6d26d903ac69c7aa58c12cR31). Every object should contain [`name` property](https://github.com/santiment/app/pull/554/files#diff-34bac4f2ff6d26d903ac69c7aa58c12cR33) that corresponds to a metric's name.
- No more `settings` since it was never used.

### Benchmarks
Benchmarks were made on the laptop in the `development` build. The results are the average of `10` runs. 
During `Preparing request` benchmark, [this section](https://github.com/santiment/app/blob/dc625cf979ed7606b3e38459c2b05d9088639f4d/src/ducks/GetTimeSeries/epics.js#L27-L75) was measured on the current branch, and [this section](https://github.com/santiment/app/blob/master/src/ducks/GetTimeSeries/epics.js#L70-L114) on the master branch.
During `Forming response` benchmark, [this section](https://github.com/santiment/app/blob/dc625cf979ed7606b3e38459c2b05d9088639f4d/src/ducks/GetTimeSeries/epics.js#L78-L83) was measured on the current branch, and [this section](https://github.com/santiment/app/blob/master/src/ducks/GetTimeSeries/epics.js#L117-L122) on the master branch.

#### First time after emptying cache and hard-reload
Old:
- Preparing request: `67,9ms`;
- Forming response: `2,1ms`;

New: 
- Preparing request: `33,5ms`;
- Forming response: `1,1ms`;

#### Following interactions
Old:
- Preparing request: `3,2ms`;
- Forming response: `1,3ms`;

New: 
- Preparing request: `2,6ms`;
- Forming response: `0,76ms`;


src/ducks/SANCharts/ChartPage.js
```diff
@@ -48,18 +46,12 @@ const LoadableChartMetricsTool = Loadable({
 
 const getChartInitialState = props => {
   let passedState
-
   if (props.location && props.location.search) {
-    const parsedQuery = mapParsedTrueFalseFields(
```
GarageInc: Why was removed? :)
Sometimes get string 'true' of string 'false' which is always boolean true in any comparisions. (for example variable 'isAdvancedView' from query)

src/ducks/SANCharts/ChartPage.js
```diff
@@ -376,13 +362,6 @@ class ChartPage extends Component {
                       />
                     </div>
                   )}
-
-                  <div className={styles.moreData}>
-                    <span className={styles.limitedLabel}>
-                      See much more data in our
-                    </span>
-                    <Link to='/dashboards'>SANbase Dashboards</Link>
```
GarageInc: Removed link to dashboards (used at /projects/bitcoin page) :( I hope I know about it.

src/pages/Chart/index.js
```diff
@@ -28,9 +28,8 @@ export default graphql(ALL_INSIGHTS_BY_PAGE_QUERY, {
   return (
     <div className={styles.wrapper + ' page'}>
       <Query query={USER_SUBSCRIPTIONS_QUERY}>
-        {({ data }) => {
-          const subscription =
-            data && getCurrentSanbaseSubscription(data.currentUser)
+        {({ data: { currentUser } = {} }) => {
+          const subscription = getCurrentSanbaseSubscription(currentUser)
```
haritonasty: sometimes backend doesn't return any data here and app is down. Will be nice to avoid it

DmitriVanGuard: Yes, it crashed because `data` was `undefined` and at the same time destructuring assignment is taken place. Here is the 3 characters fix: `= {}` - assigning default value for `data`

src/ducks/SANCharts/ChartPage.js
DmitriVanGuard: Don't worry, just a merge conflicts resolve side effect. Will be returned in the next commit

src/ducks/SANCharts/ChartPage.js
DmitriVanGuard: Can you share a screenshot/link where it's in Figma?

GarageInc: @DmitriVanGuard , here (bottom)
https://www.figma.com/file/RPLINi2nrD30hQjcANws3CxH/Sanbase?node-id=8082%3A5128

But I placed here old component(screenshot) because for new Link and upgradeBtn need to change styles of ChartWidget.js(metrics, remove old upgradeBtn in right side) - it is for another PR
  
image

But, if this component(screenshot) is not important you can remove it :)

DmitriVanGuard: `See much more data in our SANbase Dashboards` is a part of the old implementation and is not present in the up-to-date figma design (`master`)

PR #709, opened 2019-10-01
enhance/mobile-stories

haritonasty:
**Summary**
- delete arrows (changed on user clicks on content)
- enhance view, updates styles
- fix video height

**Screenshots**
image

### DIFF
--- a/src/components/Stories/StoryContent.js
+++ b/src/components/Stories/StoryContent.js
@@ -1,7 +1,9 @@
 import React, { useState, useRef } from 'react'
 import cx from 'classnames'
+import withSizes from 'react-sizes'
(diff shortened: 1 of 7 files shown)


src/components/Stories/StoryContent.js
```diff
@@ -124,7 +126,11 @@ const StoryContent = ({
               )}
               style={{
                 '--width':
-                  active === idx ? (isPaused ? `${width}px` : '100%') : 0
+                  active === idx
+                    ? isPaused
```
GarageInc: can it be simplified?

haritonasty: tried a lot, but no(
I'll play here again later

PR #718, opened 2019-10-04
Fix/signal points

GarageInc:
**Summary**

1. Fix signal triggered points. Bug with not selected DAA/Volume metrics for smart chart. Signals has wrong Y-axis position on preview. Changed vertical lines to points.
2. Add unremovable metrics for external ReferenceDots

**Screenshots**
BEFORE:
image
image

AFTER:
image
image
image


src/ducks/Signals/chart/VisualBacktestChart.js
```diff
@@ -99,24 +104,32 @@ const VisualBacktestChart = ({
...
 
-const getDataKeys = (signal = {}) => {
+export const getDataKeys = (signal = {}) => {
   if (signal.active_addresses) {
-    return { metric: 'dailyActiveAddresses', signal: 'active_addresses' }
+    return {
+      metric: 'dailyActiveAddresses',
+      key: 'daily_active_addresses',
+      signal: 'active_addresses'
+    }
```
haritonasty: Suggest to use one place of truth - Metrics

src/ducks/SANCharts/ChartActiveMetrics.js
```diff
@@ -12,6 +12,7 @@ import styles from './ChartActiveMetrics.module.scss'
 const ChartActiveMetrics = ({
   activeEvents,
   activeMetrics,
+  alwaysShowingMetrics = [],
```
haritonasty: and just for clarity - could you explain when we need this prop - `alwaysShowingMetrics`? In what cases

src/ducks/Signals/chart/VisualBacktestChart.js
GarageInc: Good point

src/ducks/SANCharts/ChartActiveMetrics.js
GarageInc: ChartWidget can't show triggered signals. They passed outside of widget as ReferenceDots with Y-position related to Y-axis of triggered metrics. And user can open this chart and unselect metric Then chart widget will remove Y-axis for this metrics. And chart will crash with error, because Reference dots has reference to removed Y-axis.
For this case I mark metric of triggered signals as 'alwaysShowing'. It mean, that only on Smart Chart of TriggerForm this metric can't me removed from chart.

GarageInc: Also we can say inside of widget that metric was removed...But it is't working, because rechart can't change reference to removed Y-axis and chart will be crashed with error to undefined Y-axis.

PR #742, opened 2019-10-18
Update all projects json

IvanIvanoff:


haritonasty:
I'm not sure, that Dima changed order just for change. I think we need to change `marketcapUsd` to `rank` and keep order by rank. It's more visual than `marketcapUsd` and, yes, shorter

IvanIvanoff:
@haritonasty if marketcap -> rank change is needed let's do it another PR. I'm updating this file from time to time otherwise new projects are not discoverable

PR #750, opened 2019-10-23
sort projects by marketcap

GarageInc:
**Summary**

Search sorted by marketcap volume

**Screenshots**
BEFORE
image

AFTER
image

### DIFF
--- a/src/components/Search/SearchContainer.js
+++ b/src/components/Search/SearchContainer.js
@@ -20,7 +20,7 @@ export const SearchContainer = ({
       className={cx(styles.wrapper, className)}
       iconPosition='left'
       onSuggestionSelect={({ slug }) => history.push(`/projects/${slug}`)}
-      sorter={({ marketcapUsd: a }, { marketcapUsd: b }) => b - a}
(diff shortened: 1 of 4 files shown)


Partysun:
@GarageInc is Marketcap always accessible here?

src/components/Search/SearchContainer.js
```diff
@@ -20,6 +20,7 @@ export const SearchContainer = ({
       className={cx(styles.wrapper, className)}
       iconPosition='left'
       onSuggestionSelect={({ slug }) => history.push(`/projects/${slug}`)}
+      sorter={({ marketcapUsd: a }, { marketcapUsd: b }) => b - a}
```
haritonasty: let's use `rank` instead `marketcapUsd` (we will have that field in our json, as I understand)

GarageInc: @Partysun , @haritonasty here are using such query
image
I will add a rank for it

GarageInc: done

PR #766, opened 2019-10-29
feature/trendings positions colors

GarageInc:
**Summary**

Add color for trending position dots

We need to show what position it was that time (more info will be on hover)
1. so red is 1-3 position in trends that day
2. orange 4-6
3. yellow 7-10

**Screenshots**
image


src/ducks/SANCharts/utils.js
```diff
@@ -390,3 +390,37 @@ export const getCrossYValue = yValue => {
...
+
+export const setColorByDayRating = events => {
+  const groups = []
+
+  events.reduce((acc, item) => {
+    const { isAnomaly, datetime } = item
+
+    if (!isAnomaly) {
+      const dateString = new Date(datetime).toDateString()
+      if (!groups[dateString]) groups[dateString] = []
```
Partysun: Don't use if condition body without "{}" pls

src/ducks/SANCharts/utils.js
```diff
@@ -390,3 +390,37 @@ export const getCrossYValue = yValue => {
...
+
+    setColor(sorted.slice(0, 3), 'var(--persimmon)')
+    setColor(sorted.slice(3, 6), 'var(--texas-rose-hover)')
+    setColor(sorted.slice(6), '#FFC24D')
+  }
+
+  return events
+}
+
+const setColor = (events, color) => {
```
Partysun: (events, color) => (events.forEach...)

src/ducks/SANCharts/utils.js
```diff
@@ -390,3 +390,37 @@ export const getCrossYValue = yValue => {
...
+    if (!isAnomaly) {
+      const dateString = new Date(datetime).toDateString()
+      if (!groups[dateString]) groups[dateString] = []
+      groups[dateString].push(item)
+    }
+
+    return groups
+  }, groups)
+
+  for (let key in groups) {
```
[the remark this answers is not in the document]
GarageInc: Correct, will do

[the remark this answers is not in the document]
GarageInc: No, can't do. We have a list of not sorted by value items for days range.
First step: group them by days(in reduce-function). One day got one list of items. We do't know that we got a list of all items for selected day before looping over all items.
Second step: sort every list by position/value. The top items get color 'persimon', middle itmes get second color, and last items get a third color.  @Partysun

src/ducks/SANCharts/Charts.js
```diff
@@ -456,6 +457,8 @@ class Charts extends React.Component {
         metrics.some(({ key }) => key === value) || !isAnomaly
     )
 
+    events = setColorByDayRating(events)
```
DmitriVanGuard: No point for making `n` more iterations and mutating the object, when a correct color can be assigned inside loop on the `LOC 612`

src/ducks/SANCharts/Charts.js
```diff
@@ -606,14 +609,14 @@ class Charts extends React.Component {
...
-              events.map(({ key, y, datetime }) => (
+              events.map(({ key, y, datetime, color = 'var(--persimmon)' }) => (
                 <ReferenceDot
                   yAxisId={`axis-${key}`}
                   r={3}
                   isFront
                   fill='var(--white)'
                   strokeWidth='2px'
-                  stroke='var(--persimmon)'
+                  stroke={color}
```
DmitriVanGuard:
```js
const getTrendRatingColor = value => {
    if(value < 4) return 'var(--persimmon)';
    if(value < 7) return 'var(--texas-rose-hover)';
    return 'var(--bright-sun)';
}

< /*...*/
stroke={isAnomaly ? color : getTrendRatingColor(value)}
/*...*/ />
```

src/ducks/SANCharts/utils.js
```diff
@@ -390,3 +390,25 @@ export const getCrossYValue = yValue => {
...
+        item.color = 'var(--persimmon)'
+      } else if (value <= 6) {
+        item.color = 'var(--texas-rose-hover)'
+      } else {
+        item.color = 'var(--bright-sun)'
+      }
+    }
+
+    return groups
+  }, groups)
```
DmitriVanGuard: Why `groups` is the second argument?

src/ducks/SANCharts/Charts.js
```diff
@@ -237,12 +238,12 @@ class Charts extends React.Component {
         const { metricAnomalyKey: anomaly } = rest
         const result = value || chartData[index]
         const eventsData = getEventsTooltipInfo(rest)
```
haritonasty: for me better to move color calculation to [`getEventsTooltipInfo`](https://github.com/santiment/app/blob/master/src/ducks/SANCharts/utils.js#L15) function

PR #769, opened 2019-10-30
feature/chart above below signals

GarageInc:
**Summary**

Add support of above/below price signals on charts (comparing with current price)

**Screenshots**

image
image


src/ducks/SANCharts/Charts.js
```diff
@@ -145,6 +139,13 @@ class Charts extends React.Component {
     })
   }
 
+  buildChartSignalData = (chartY, priceUsd, signal = undefined) => ({
```
haritonasty: you can skip `undefined`

PR #777, opened 2019-10-31
feature/metrics-divider

DmitriVanGuard:
### Summary
Dividing selected metrics into separated charts
### Screenshots
gif


src/ducks/SANCharts/Charts.js
```diff
@@ -177,6 +181,51 @@ class Charts extends React.Component {
...
+            isDayStartMetric(
+              data,
+              dayMetric,
+              dayStartMetrics,
+              currentPointDatetime
+            )
+          ) {
+            continue
+          }
+          assignToPointDayStartValue(
```
GarageInc: should be moved inside if(!....) block with removing 'continue'

src/ducks/SANCharts/Charts.js
```diff
@@ -177,6 +181,51 @@ class Charts extends React.Component {
...
+        ({ key, minInterval }, index) =>
+          minInterval === '1d' && dayMetrics.push([key, index])
+      )
+
+      const dayStartMetrics = {}
+
+      alignedChartData = chartData.map(({ ...data }) => {
+        const { datetime: currentPointDatetime } = data
+
+        for (let i = 0; i < dayMetrics.length; i++) {
```
GarageInc: foreach? 'i' is not using inside loop

DmitriVanGuard: 1. `LOC 200` -> `dayMetrics[i]`
2. Creating functions inside the loop is a bad idea. `chartData` usually has `150+` items, which will result in the same amount of **a new closures**

PR #826, opened 2019-11-26
enhance/payment-modal

DmitriVanGuard:
### Summary
2 columns checkout form layout.

### Screenshots
image


Partysun:
iOS Safari has limited FormData API subfeatures, I don't check what subfeature did we use. Question is it ok?
@DmitriVanGuard

PR #883, opened 2019-12-25
enhance/mobile-chart-tooltips

haritonasty:
**Summary**
change tooltip styles via new design

**Screenshot**


src/ducks/SANCharts/utils.js
```diff
@@ -237,7 +239,7 @@ export const generateMetricsMarkup = (
         key={`line-${dataKey}`}
         type='linear'
         yAxisId={currentYAxisId}
-        name={label}
+        name={(useShortName && shortLabel) || label}
```
Partysun: name={ use ? shortL : la } @haritonasty

haritonasty: ternary here not the same case: not every metric has `shortLabel`

PR #927, opened 2020-01-17
Add gtag config

Partysun:
- Switch from analytics.js to gtag.js
- Remove react-ga
- Keep interface of .event method which we used in the app
- GA instead of GoogleAnalytics everywhere
- Load gtag and fire any event only if do not track is false


src/utils/tracking.js
```diff
@@ -0,0 +1,104 @@
...
+  importGTAG.async = true
+  importGTAG.type = 'text/javascript'
+  importGTAG.src = '//www.googletagmanager.com/gtag/js?id=' + TRACKER_IDs[0]
+
+  const head = document.getElementsByTagName('head')[0]
+  head.appendChild(importGTAG)
+}
+
+export function initializeTracking (trackerIDs = TRACKER_IDs) {
+  if (isBrowser && process.env.BACKEND_URL === PRODUCTION_API_URL) {
```
GarageInc: Maybe need to use 'process.env.NODE_ENV === 'production'?

Partysun: @GarageInc stage server is production also.

PR #967, opened 2020-01-30
feature/svelte-components

DmitriVanGuard:
## Summary
- Making possible to use `svelte` components in react;
   - `webpack config` was extended (without ejecting) in order to import `.svelte` files. 
   - `scripts/index.js` is necessary to prepare `insights-app` for usage as library;
   - Global styles were added to preserve compatibility.
- `Comments` modal for insights card.
## Example 
```jsx
import React from 'react'
import toReact from 'svelte-adapter/react'
import SvelteComments from 'svelte-ui/Comments.svelte'

const Comments = toReact(SvelteComments, null, 'div')

export default () => <Comments />

```

## Screenshots
gif


.gitignore
```diff
@@ -1,6 +1,7 @@
 # See https://help.github.com/ignore-files/ for more about ignoring files.
 /node_modules
 /build
+undefined.json
```
Partysun: What is this? @DmitriVanGuard

PR #1079, opened 2020-03-18
migrate/new-font

haritonasty:
**Screenshots**
image
image
image
image
image


Partysun:
Looks pretty simple switch. @haritonasty :)

haritonasty:
@Partysun because designers said that we need to change only buttons, so didn't update all illustrations. Font looks ok for me everywhere, some places where we set font-size manually (without mixin) we can change step by step.

PR #1080, opened 2020-03-19
enhance/social-tool

haritonasty:
**Summary**
New layout for `/trends/explore/` page: 
- add sidebar with wordcloud and trends table, 
- fix main chart via new design (later will migrate to `san-charts`)

**Screenshot**
image


src/components/Trends/Explore/TrendsExploreAdditionalInfo.js
```diff
@@ -14,7 +10,7 @@ const NEWS_INDEX = 'News'
 const INSIGHTS_INDEX = 'Insights'
 
 const TrendsExploreAdditionalInfo = ({
-  news: newsRaw,
+  news: newsRaw = [],
```
Partysun: @haritonasty We can remove all "news" related code

PR #1086, opened 2020-03-20
metrics insights component

GarageInc:
**Summary**

Show insights for some metrics

**Screenshots**

image


src/ducks/Studio/Chart/MetricsExplanation/MetricInsight/MetricInsightCard.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
```
Partysun: @GarageInc This component better to move in components folder. It can be used in other places + it depends on components.

src/ducks/Studio/Chart/MetricsExplanation/MetricInsight/MetricInsights.js
```diff
@@ -0,0 +1,36 @@
...
+    return null
+  }
+
+  return (
+    <>
+      <div className={styles.usecases}>Use cases from Insights</div>
+      <div className={styles.container}>
+        {insights.map(id => {
+          return (
+            <Query query={INSIGHT_BY_ID_QUERY} variables={{ id }} key={id}>
```
Partysun: Why do we need to load anything? @GarageInc

GarageInc: @Partysun , how will we know about author and name of insight without loading? Or does it need to hardcode such data? (user can change title of insight or his username)

Partysun: @GarageInc A. Ok. Let's fetch this info.

PR #1090, opened 2020-03-24
add comments for feed timeline events

GarageInc:
**Summary**

Add comments for all feed cards

**Screenshots**

image
image


src/components/Insight/Comments.js
```diff
@@ -58,24 +63,44 @@ function editComment (id, content) {
...
+  onDelete = deleteComment,
+  onCreate = createComment,
+  onGet = getComments
+}) => {
   const [comments, setComments] = useState([])
   const [loading, setLoading] = useState(true)
 
-  useEffect(() => {
-    getComments(id).then(({ data }) => {
+  const loadCommentsCount = () => {
```
DmitriVanGuard: `loadCommentsCount` is a misleading name, it implies the process of getting a single value – **comments count**. Meanwhile, array of comments is being fetched.

src/components/Insight/Comments.js
```diff
@@ -58,24 +63,44 @@ function editComment (id, content) {
...
+
+  useEffect(() => {
+    loadCommentsCount()
   }, [])
 
+  const onClose = () => {
+    loadCommentsCount()
+  }
+
+  let showingCount = comments.length || count
```
DmitriVanGuard: 1. Value is never reassigned;
2. Better to call this variable `commentsCount`;
3. Should be initialised at the top of comment, otherwise it's lost between all hooks definition and `return` statement.

src/components/Insight/Comments.js
```diff
@@ -58,24 +63,44 @@ function editComment (id, content) {
...
       setLoading(false)
     })
+  }
+
+  useEffect(() => {
+    loadCommentsCount()
   }, [])
 
+  const onClose = () => {
+    loadCommentsCount()
```
DmitriVanGuard: `loadCommentsCount` can be passed directly as the reference, since it's the only function call

src/components/Insight/Comments.js
```diff
@@ -58,24 +63,44 @@ function editComment (id, content) {
...
   return (
     <Modal
       trigger={
         <div className={cx(sharedStyles.stat, sharedStyles.stat_comments)}>
-          <Icon type='comment' className={sharedStyles.commentIcon} /> {count}
+          <Icon type='comment' className={sharedStyles.commentIcon} />{' '}
+          {showingCount}
         </div>
       }
+      onClose={onClose}
```
DmitriVanGuard: `onOpen` will guarantee up-to-date state, `onClose` won't

src/components/Insight/Comments.js
```diff
@@ -7,34 +7,39 @@ import toReact from 'svelte-adapter/react'
 import SvelteComments from 'insights-app/lib/components/comments/Comments'
 import { client } from '../../index.js'
 import {
```
DmitriVanGuard: **[D-0-0]** Since this component is used as a generic interface for working with comments, `src/components/Insight/Comments.js` component should be abstracted away from any target realisation, i.e no default arguments related to the specific case. `editComment`, `deleteComment`,`createComment`,`getComments` should be passed only as the props when using it. By doing that, component's internal design will be significantly improved

src/components/Insight/Comments.js
```diff
@@ -58,24 +63,44 @@ function editComment (id, content) {
...
 
-export default ({ id, authorId, count }) => {
+export default ({
+  id,
+  authorId,
+  count,
+  onEdit = editComment,
+  onDelete = deleteComment,
+  onCreate = createComment,
+  onGet = getComments
```
DmitriVanGuard: `onEdit `, `onDelete`, ...etc are misleading function names. Function following this convention implies that function is a callback, which will be called after specific event occurred. This is clearly represented by the browser native events: `onclick`, `onkeypress`, etc. After making an according changes to the **[D-0-0]** comment, this issue will be resolved

PR #1095, opened 2020-03-27
enhance/price-histogram

DmitriVanGuard:
## Summary
- Displaying circle histogram;
- Showing current project's price near the related bucket;
- Circles after the bucket containing the current price are marked with a red color;

## Screenshots
image


src/ducks/Studio/AdvancedView/Histogram/index.js
```diff
@@ -31,13 +33,21 @@ function rangeFormatter ([left, right]) {
   return usdFormatter(left) + ' - ' + usdFormatter(right)
 }
 
-const Frame = ({ range, value, ticker, width }) => {
+const Frame = ({ range, value, ticker, width, price, isRed }) => {
```
Partysun: @DmitriVanGuard Millify we are doing inside this func, why than we can't calculate red color using price and range fields. And make less interface.

src/ducks/Studio/AdvancedView/Histogram/hooks.js
```diff
@@ -1,14 +1,14 @@
...
-  height: 342,
+  height: 50,
   top: 0
 }
 
-function formatHistogramData (data) {
+function formatHistogramData (data, price) {
   const { length } = data
 
   let max = 0
```
Partysun: I am thinking the size of data is predictable and small no more than 20.
Maybe better to use one liner to make less code and inc readability.
Math.max(...data.map(el=>el.value))
? @DmitriVanGuard

src/ducks/Studio/AdvancedView/Histogram/index.js
```diff
@@ -72,10 +82,12 @@ const Histogram = ({ title, slug, ticker, date, datesRange, hasSort }) => {
     setDates(newDates)
   }
 
+  let isBucketAfterCurrentPrice = false
+
   return (
     <div className={styles.wrapper}>
       <h2 className={styles.title}>
-        Price movement
+        Price distribution
```
Partysun: Spent coin cost

src/ducks/Studio/AdvancedView/Histogram/hooks.js
DmitriVanGuard: I'm mean, this is a 3 line algorithm (loop, value check, assignment), which intention is pointed out beforehand with a variable called `max`.

The difference in speed for the array with 10 items: image
For a simple task, it's a huge difference.

src/ducks/Studio/AdvancedView/Histogram/index.js
DmitriVanGuard: Addressed issue was fixed in the new commits

src/ducks/Studio/AdvancedView/Histogram/index.js
DmitriVanGuard: Addressed issue was fixed in the new commits

PR #1096, opened 2020-03-27
Add Gold & SPX prices for comparison

Partysun:
image

### DIFF
--- a/src/ducks/Studio/Compare/index.js
+++ b/src/ducks/Studio/Compare/index.js
@@ -11,6 +11,11 @@ import { MAX_METRICS_AMOUNT } from '../constraints'
 import { useChartColors } from '../../Chart/colors'
 import styles from './index.module.scss'
 
+const FiatMarketAssets = [
+  { slug: 's-and-p-500', name: 'S&P500', ticker: 'SPX' },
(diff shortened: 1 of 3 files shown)


src/ducks/Studio/Compare/index.js
```diff
@@ -24,8 +24,15 @@ const Compare = ({
...
+      const mixedProjects = [
+        ...allProjects,
+        { slug: 's-and-p-500', name: 'S&P500', ticker: 'SPX' },
+        { slug: 'gold', name: 'Gold', ticker: 'Gold' }
+      ]
       setProjects(
-        allProjects.filter(project => project.slug !== slug).sort(projectSorter)
+        mixedProjects
+          .filter(project => project.slug !== slug)
+          .sort(projectSorter)
```
DmitriVanGuard: `SPX` and `Gold` could be moved outside of the function as a constant array, e.g. `const ADDITIONAL_PROJECTS = [{}, {}, ...]`, and appended to the `allProjects` via `allProjects.concat(ADDITIONAL_PROJECTS)`. This will give us a benefit in speed since no new objects will be created on function call, and `concat` is slightly faster for array concatenations

Partysun: Thx. concat looks better.

PR #1109, opened 2020-04-06
add signals for social volume and other metrics

GarageInc:
**Summary**

Add new signals(by metrics) and add them into chart

**Screenshots**

image
image


src/ducks/Studio/Alerts/suggestions/index.js
```diff
@@ -9,5 +14,26 @@ export const Suggestion = Object.assign(Object.create(null), {
   daily_active_addresses: {
     title: 'DAA',
     suggesters: dailyActiveAddressesSuggesters
+  },
+  social_volume_total: {
+    title: 'Social Volume',
+    suggesters: [
+      suggestValueChange(Metric.social_volume_total),
+      suggestPercentUpDown(Metric.social_volume_total)
```
DmitriVanGuard, line 22: Since this is a generic pattern, these functions should be moved outside of the `suggestions/price` module to `suggestions/helpers`

src/ducks/Studio/Alerts/suggestions/price.js
```diff
@@ -3,36 +3,43 @@ import { createSuggestion } from './helpers'
...
 
-const { formatter } = Metric.price_usd
+export const SIGNAL_BELOW = 'BELOW'
+export const SIGNAL_ABOVE = 'ABOVE'
+export const VALUE_IFS = ['drops below', 'rises above']
 
-const SIGNAL_BELOW = 'BELOW'
-const SIGNAL_ABOVE = 'ABOVE'
-const PRICE_IFS = ['drops below', 'rises above']
+export const suggestValueChange = metric => {
```
DmitriVanGuard, line 15: Now the name`suggestValueChange` breaks the semantic of the function, because it no longer returns a **suggestion**, but a suggester (a function), that will return a suggestion. My suggestion is to rename it to something like `buildValueChangeSuggester`

src/ducks/Chart/Signals/helpers.js
```diff
@@ -134,6 +137,17 @@ export function makeSignalDrawable (
...
+  daily_active_addresses: buildDAASignal,
+  social_volume_total: (slug, value, type) => {
+    return buildValueChangeSignal(slug, value, type, Metric.social_volume_total)
+  },
+  exchange_balance: (slug, value, type) => {
+    return buildValueChangeSignal(slug, value, type, Metric.exchange_balance)
+  },
+  transaction_volume: (slug, value, type) => {
+    return buildValueChangeSignal(slug, value, type, Metric.transaction_volume)
+  }
```
DmitriVanGuard, line 152: Metric could be passed as 4th argument to the mapped function here:
https://github.com/santiment/app/blob/e0ea4c6c38f85885f393e068534f908bb612a04c/src/ducks/Chart/Signals/index.js#L105
This will allow to assign `buildValueChangeSignal` as a reference and reduce repetitive code.

PR #1162, opened 2020-04-28
validate metrics interval if DAA in list

GarageInc:
**Summary**

Bug: showed empty bars for DAA if we have a lot of metrics with small intervals(for example '12h'). It depended from loaded data queue

Hotfix: make strict '1d' interval for all metrics if we have DAA on list


src/ducks/Studio/timeseries/hooks.js
```diff
@@ -3,6 +3,7 @@ import { client } from '../../../index'
 import { getQuery, getPreTransform } from './fetcher'
 import { normalizeDatetimes } from './utils'
 import { mergeTimeseriesByKey } from '../../../utils/utils'
+import {Metrics} from "../../SANCharts/data";
```
haritonasty, line 6: why from SANCharts? new place - https://github.com/santiment/app/blob/master/src/ducks/dataHub/metrics/index.js#L81

haritonasty: Could you pls remove `;` at the end of row

GarageInc: IDEA :(

Replaced

haritonasty: change settings in IDEA for this case)

Partysun: image
Honestly also "Metrics" should have space around.

src/ducks/Studio/timeseries/hooks.js
```diff
@@ -124,12 +127,14 @@ export function useTimeseries (
...
 
+        const strictInterval = hasDaaMetric ? '1d' : interval
+
         client
           .query({
             query,
             variables: {
               metric: key,
-              interval,
+              interval: strictInterval,
```
DmitriVanGuard, line 137: `useTimeseries` is and should be only responsible for fetching and aggregating data. It has a public interface for tweaking metric-specific query arguments – `MetricSettingMap` (third argument)

src/ducks/Studio/timeseries/hooks.js
```diff
@@ -97,6 +98,8 @@ export function useTimeseries (
       let raceCondition = false
       let mergedData = []
 
+      const hasDaaMetric = metrics.some(({key}) => key === Metric.daily_active_addresses.key)
```
DmitriVanGuard, line 101: Since `metrics` is an array, `.includes` is a better way to check existence of the element inside the array.

src/ducks/Studio/timeseries/hooks.js
GarageInc: Command 'yarn format' was removed 2 weeks ago. Returned.

src/ducks/Studio/timeseries/hooks.js
GarageInc: @DmitriVanGuard , :(
A lot of places with usage of useTimeseries. Programmer can forgor about DAA interval if he will add any chart with this hook.
image

GarageInc: Changed(CUSTOM_METRIC_SETTINGS_MAP), need check

src/ducks/Studio/index.js
```diff
@@ -40,44 +40,46 @@ const Studio = ({
...
   const [advancedView, setAdvancedView] = useState()
   const [shareLink, setShareLink] = useState()
   const [isICOPriceDisabled, setIsICOPriceDisabled] = useState(true)
-  const [isSidebarClosed, setIsSidebarClosed] = useState()
+
   const [rawData, loadings, ErrorMsg] = useTimeseries(
     activeMetrics,
     settings,
-    MetricSettingMap,
+    getPreparedMetricSettings(activeMetrics, MetricSettingMap)
```
DmitriVanGuard, line 53: `getPreparedMetricSettings` will be called on every render, which is bad. It should be a hook, which depends on the `activeMetrics` and `MetricSettingMap`

src/ducks/Studio/timeseries/hooks.js
```diff
@@ -56,6 +57,30 @@ function abortAllMetrics (abortables) {
...
+
+      delete newValue['delete']
+
+      settings.set(key, newValue)
+    })
+  }
+
+  return settings
+}
+
```
DmitriVanGuard, line 83: 1. This file is for hooks or functions that are directly related to the hook (e.g. `abortRemovedMetrics`);
2. Since `settings` is a map, variable name should be more descriptive, e.g `settingsMap`;
3. Better to return a new map, not to mutate the old one.

src/ducks/Studio/timeseries/hooks.js
```diff
@@ -56,6 +57,30 @@ function abortAllMetrics (abortables) {
...
+        interval: '1d'
+      })
+    })
+  } else {
+    settings.forEach((value, key) => {
+      const newValue = {
+        ...value
+      }
+
+      delete newValue['delete']
```
DmitriVanGuard, line 75: @GarageInc what is the reason for deleting the `delete` property?

GarageInc: My mistake

GarageInc:
@DmitriVanGuard , check please

PR #1163, opened 2020-04-28
iterate over timeseries by date difference

GarageInc:
**Summary**
Bug: showed empty bars for DAA if we have a lot of metrics with small intervals(for example '12h'). It depended from loaded data queue

Hotfix: calculate date difference (not lenght of timeseries arrays)


DmitriVanGuard:
Closing in favor of #1169

PR #1241, opened 2020-06-03
Feature/alpha block

GarageInc:
**Summary**
Add alpha block on index page. Link to sheets or to pricing (if not pro user)

**Screenshots**

image
image

### DIFF
--- a/src/utils/account.js
+++ b/src/utils/account.js
@@ -3,10 +3,11 @@ import { PRO } from '../components/Navbar/NavbarProfileDropdown'
 
 export const checkIsProUser = data => {
   const sanbaseSubscription = getCurrentSanbaseSubscription(data)
-
-  return sanbaseSubscription && sanbaseSubscription.plan
-    ? sanbaseSubscription.plan.name === PRO
-    : false
+  return (
+    sanbaseSubscription &&
+    sanbaseSubscription.plan &&
+    sanbaseSubscription.plan.name === PRO
+  )
 }
 
 export const checkIsProState = ({ user: { data } }) => {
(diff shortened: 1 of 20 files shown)


Partysun:
Take this onhold before final flow will be ready by designers. @GarageInc

src/utils/account.js
```diff
@@ -0,0 +1,16 @@
+import { getCurrentSanbaseSubscription } from './plans'
+import { PRO } from '../components/Navbar/NavbarProfileDropdown'
+
+export const checkIsProUser = data => {
+  const sanbaseSubscription = getCurrentSanbaseSubscription(data)
+
+  return sanbaseSubscription && sanbaseSubscription.plan
+    ? sanbaseSubscription.plan.name === PRO
```
Partysun, line 8: @GarageInc why not && sanbase.plan.name === PRO

src/utils/account.js
```diff
@@ -0,0 +1,16 @@
...
+    : false
+}
+
+export const checkIsProState = ({ user: { data } }) => {
```
Partysun, line 12: It can be just "isUserPro" or "isPro" @GarageInc

Partysun: State is 100% must be removed

GarageInc: Agree. But it is only simplified code. I imported copy paste from project in one place with 2 functions, because in some places we are checking state, in other places we are checking only user.

src/utils/account.js
GarageInc: 'copypaste'
Will change

PR #1754, opened 2020-11-02
fix/btc-dashboard/skeleton-flashing

haritonasty:


haritonasty:
For Skeleton component need to have wrapper and concrete height to prevent covering for whole page, @GarageInc

PR #2143, opened 2021-04-26
enhance/home-page

GarageInc:
## Changes

Changed home page components

## Notion's card

https://www.notion.so/santiment/Reorder-recently-viewed-on-home-page-e44495f4a4c143259a438ea6062d11c2

## Checklist

- [x] I've performed a self-review, followed all rules from [Frontend style guide](https://www.notion.so/santiment/Front-end-style-guide-81750096b38c4bea9a29b14fd4ab8667)
- [x] If I make changes to another person's module, I've asked how to use it or request a review
- [x] I've updated the [documentation](https://github.com/santiment/academy), if necessary (Keyboard shortcuts, Account settings)
- [x] I've checked night mode, mobile & tablet screens (if have changes in UI)
- [x] I've added tests (if necessary)

## Screenshots or GIFs

image

### DIFF
--- a/src/components/SignalCard/card/creator/SignalCreator.js
+++ b/src/components/SignalCard/card/creator/SignalCreator.js
@@ -22,7 +22,11 @@ const SignalCreator = ({
   const nameOrEmail = username || email
 
   return (
-    <Link to={'/profile/' + id} className={cx(styles.container, className)}>
+    <Link
+      to={'/profile/' + id}
+      className={cx(styles.container, className)}
+      onClick={e => e.stopPropagation()}
+    >
       <UserAvatar
         as='div'
         userId={id}
--- a/src/components/SignalCard/card/creator/SignalCreator.js
+++ b/src/components/SignalCard/card/creator/SignalCreator.js
@@ -8,6 +8,8 @@ import styles from './SignalCreator.module.scss'
 export const showUserActions = () =>
   window && window.location && window.location.pathname !== personalLocation
 
+const onClick = e => e.stopPropagation()
+
 const SignalCreator = ({
   className,
   classes = {},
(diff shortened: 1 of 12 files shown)


src/components/SignalCard/card/creator/SignalCreator.js
```diff
@@ -22,7 +22,11 @@ const SignalCreator = ({
   const nameOrEmail = username || email
 
   return (
-    <Link to={'/profile/' + id} className={cx(styles.container, className)}>
+    <Link
+      to={'/profile/' + id}
+      className={cx(styles.container, className)}
+      onClick={e => e.stopPropagation()}
```
haritonasty, line 28: better to move this function in onClick outside to avoid recreations

src/pages/Index/Aside/Recents/index.js
```diff
@@ -32,12 +50,26 @@ const Recents = () => {
...
+      />
+
+      {Component && (
+        <Component
+          assets={assets}
+          chartLayouts={chartLayouts}
+          watchlists={watchlists}
+          screeners={screeners}
+        />
+      )}
```
haritonasty, line 72: I am not sure but looks like even if a component doesn't contain items (for ex. no recent screeners), the tab will appear and will be just empty. Did you discuss with Anastasya that case and what we should show?

haritonasty: Also here every component will receive items from other components, which won't use.

src/components/SignalCard/card/creator/SignalCreator.js
GarageInc: agree

src/pages/Index/Aside/Recents/index.js
GarageInc: Ok, will check such case

PR #2496, opened 2022-05-17
Unused Navbar code removed

mqklin:


DmitriVanGuard:
> @DmitriVanGuard Do we use this Navigation code or we are using web-kit?

Yes, we are reusing the `san-webkit` for product links, which has the `sanr.app` as the SanR link.
This Navbar code is unused part of the codebase. I have made the commits to remove it.

